PR #4, opened 2020-03-20
Migration to add table for storing form responses

brucebolt:
We need to store form responses from users.  Adding a basic migration to store the submission timestamps and form response (as a JSON representation).  Using JSON will allow the questions to remain fluid without needing to change the database schema.

Trello card: https://trello.com/c/4tLmMu1n

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -1,5 +1,3 @@
-# frozen_string_literal: true
-
 # This file is auto-generated from the current state of the database. Instead
 # of editing this file, please use the migrations feature of Active Record to
 # incrementally modify your database, and then regenerate this schema definition.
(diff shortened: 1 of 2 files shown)


db/schema.rb
```diff
@@ -12,7 +10,14 @@
...
 
-ActiveRecord::Schema.define(version: 0) do
+ActiveRecord::Schema.define(version: 2020_03_20_105826) do
+
   # These are extensions that must be enabled in order to support this database
-  enable_extension 'plpgsql'
+  enable_extension "plpgsql"
+
+  create_table "form_responses", force: :cascade do |t|
+    t.json "form_response", default: {}, null: false
```
richardTowers: Can we get this to use `jsonb` instead of `json`?

richardTowers: Maybe it already does, I don't know rails lol

brucebolt: Sure 👍

brucebolt: Done.  It won't use jsonb unless we explicitly state it.

richardTowers (approved):

PR #11, opened 2020-03-20
Add FormResponse model

brucebolt:
This will be used to store user responses when the form is submitted.

Trello card: https://trello.com/c/4tLmMu1n


richardTowers:
Do we want one model for the whole application? Or one model per page?

brucebolt:
I was thinking of one model that will store the whole response.  We can write this only when the final page of the form is submitted.

bilbof (approved):

PR #14, opened 2020-03-20
Add question 1.1: Medical equipment

bilbof:
This also adds stubs for 1.1 and 2

### DIFF
--- a/app/controllers/coronavirus_form/medical_equipment_controller.rb
+++ b/app/controllers/coronavirus_form/medical_equipment_controller.rb
@@ -1,0 +1,40 @@
+# frozen_string_literal: true
+
+class CoronavirusForm::MedicalEquipmentController < ApplicationController
+  include ActionView::Helpers::SanitizeHelper
+  include FieldValidationHelper
+
+  def show
+    render "coronavirus_form/#{PAGE}"
+  end
+
+  def submit
+    medical_equipment = sanitize(params[:medical_equipment]).presence
+
+    session[:medical_equipment] = medical_equipment
+
+    invalid_fields = validate_radio_field(
+      PAGE,
+      radio: medical_equipment,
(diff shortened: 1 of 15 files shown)


app/controllers/coronavirus_form/medical_equipment_controller.rb
```diff
@@ -0,0 +1,40 @@
...
+      allowed_values: I18n.t("coronavirus_form.#{PAGE}.options").map { |_, item| item.dig(:label) }
+    )
+
+    if invalid_fields.any?
+      flash.now[:validation] = invalid_fields
+      render "coronavirus_form/#{PAGE}"
+    elsif has_seen_check_answers?
+      redirect_to controller: 'coronavirus_form/check_answers', action: 'show'
+    else
+      controller = medical_equipment == "Yes" ? "coronavirus_form/medical_equipment_kind" : "coronavirus_form/hotel_rooms"
```
brucebolt: I'd suggest using the translation value here rather than hardcoding 'Yes', just in case these values change

spec/controllers/coronavirus-form/medical_equipment_controller_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+  describe 'POST submit' do
+    let(:selected) { "Yes" }
+
+    it 'sets session variables' do
+      post :submit, params: { medical_equipment: selected }
+
+      expect(session[session_key]).to eq selected
+    end
+
+    it 'redirects to medical equipment kind' do
```
brucebolt:
```suggestion
    it 'redirects to medical equipment kind when yes selected' do
```

Or maybe wrap in a context block, like for 'no'?

brucebolt (commented):
Looks good, I've just added a few comments

PR #15, opened 2020-03-20
Add "Do you have medical equipment to offer?" question

brucebolt:
This question is the first in the series of questions.  It has logic where a 'yes' response send the user through a further series of sub-questions about medical equipment.  If 'no' is selected, they are taken to the next question.

Screenshot 2020-03-20 at 15 42 40

Trello card: https://trello.com/c/3utkqx5M

### DIFF
--- a/app/controllers/coronavirus_form/medical_equipment_controller.rb
+++ b/app/controllers/coronavirus_form/medical_equipment_controller.rb
@@ -1,0 +1,40 @@
+# frozen_string_literal: true
+
+class CoronavirusForm::MedicalEquipmentController < ApplicationController
+  include ActionView::Helpers::SanitizeHelper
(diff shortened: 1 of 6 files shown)


app/controllers/coronavirus_form/medical_equipment_controller.rb
```diff
@@ -0,0 +1,41 @@
+# frozen_string_literal: true
+
+class CoronavirusForm::MedicalEquipmentController < ApplicationController
+  include ActionView::Helpers::SanitizeHelper
+  include FieldValidationHelper
+
+  def show
+    session[:medical_equipment] ||= []
```
bilbof: This isn't necessary.

bilbof (approved):
We also need to update the start page

PR #16, opened 2020-03-20
Load spec_helper in test

leenagupte:
We're not using rails, so this was failing


bilbof (approved):

issyl0:
Hmm, looks like a Rails app to me: https://github.com/alphagov/govuk-coronavirus-business-volunteer-form/blob/master/Gemfile#L7. 🤔

bilbof:
It was failing because that file didn't exist, I think because we're using rspec-rails.

PR #17, opened 2020-03-20
Import RuboCop config from rubocop-govuk and lint all the things

issyl0:
- This at least starts us off with a passing build.
- The scss-lint part of the `lint` Rake task doesn't currently work,
  which means that the `lint` Rake task doesn't currently work.


huwd (approved):
just that one conflict to resolve but looks great

issyl0:
Resolved. Thanks Huw!

PR #22, opened 2020-03-21
Get possible responses from i18n file (for question 1)

brucebolt:
Removing the hardcoding of specific response options and simply loading all the options included in the i18n YAML file for this question will make it simpler to change the questions/answers.

### DIFF
--- a/app/views/coronavirus_form/medical_equipment.html.erb
+++ b/app/views/coronavirus_form/medical_equipment.html.erb
@@ -17,18 +17,13 @@
   heading: t('coronavirus_form.medical_equipment.title'),
   is_page_heading: true,
   name: "medical_equipment",
-  items: [
+  items: t('coronavirus_form.medical_equipment.options').map do |_, item|
     {
-      value: t('coronavirus_form.medical_equipment.options.option_yes.label'),
-      text: t('coronavirus_form.medical_equipment.options.option_yes.label'),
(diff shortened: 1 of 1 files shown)


app/views/coronavirus_form/medical_equipment.html.erb
```diff
@@ -17,18 +17,13 @@
...
-    },
-    {
-      value: t('coronavirus_form.medical_equipment.options.option_no.label'),
-      text: t('coronavirus_form.medical_equipment.options.option_no.label'),
-      checked: session[:medical_equipment] == t('coronavirus_form.medical_equipment.options.option_no.label'),
-    },
-  ]
+      value: item[:label],
+      text: item[:label],
+      checked: session[:hotel_rooms] == item[:label],
```
issyl0:
```suggestion
      checked: session[:medical_equipment] == item[:label],
```

:question:

brucebolt: Oops!  Copy and pasted this from another view.

issyl0: This is why we have code review especially at this time of day. :laughing:

issyl0 (approved):

app/views/coronavirus_form/medical_equipment.html.erb
brucebolt: Couldn't agree more!

PR #27, opened 2020-03-21
Ship logs to splunk

richardTowers:
I created the splunk ssl drain service by hand, using:

```
cf create-service splunk unlimited splunk-ssl-drain
```

(Having previously enabled the splunk service broker in the
govuk_development org)


tlwr (approved):
`splunk-ssl-drain` exists in the relevant space

PR #28, opened 2020-03-21
Rename coronavirus-form spec directory

brucebolt:
The controller spec directory is named as `coronavirus-form`, but the corresponding application controller directory is named `coronavirus_form`.  Renaming to keep this consistent.


huwd (approved):

brucebolt:
I've rebased against master, since new specs have recently been added

PR #31, opened 2020-03-21
Implement question 5 (Can you offer space?)

brucebolt:
Adds an additional question and all the relevant logic to move to the next/previous questions.

Also introduces a placeholder route for the next question.

Screenshot 2020-03-21 at 09 57 09

### DIFF
--- a/app/controllers/coronavirus_form/offer_space_controller.rb
+++ b/app/controllers/coronavirus_form/offer_space_controller.rb
@@ -1,0 +1,37 @@
+# frozen_string_literal: true
+
+class CoronavirusForm::OfferSpaceController < ApplicationController
+  include ActionView::Helpers::SanitizeHelper
+  include FieldValidationHelper
+
+  def show
+    render "coronavirus_form/#{PAGE}"
+  end
+
+  def submit
+    offer_space = sanitize(params[:offer_space]).presence
+    session[:offer_space] = offer_space
+
+    invalid_fields = validate_radio_field(
+      PAGE,
+      radio: offer_space,
+    )
+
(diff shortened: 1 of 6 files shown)


app/controllers/coronavirus_form/offer_space_controller.rb
```diff
@@ -0,0 +1,37 @@
...
+      radio: offer_space,
+    )
+
+    if invalid_fields.any?
+      flash.now[:validation] = invalid_fields
+      render "coronavirus_form/#{PAGE}"
+    elsif session["check_answers_seen"]
+      redirect_to controller: "coronavirus_form/check_answers", action: "show"
+    else
+      redirect_to controller: "coronavirus_form/offer_advice", action: "show"
```
bilbof, line 26: This will need to be changed to the 'what kind of space' controller

richardTowers (approved):

bilbof (approved):
Looks good, section 5.1 has been added since but that could come in another PR

app/controllers/coronavirus_form/offer_space_controller.rb
issyl0: Ah, you chose "offer advice". I went with "expert advice"?

brucebolt:
Great.  I'll write a card to implement 5.1.

app/controllers/coronavirus_form/offer_space_controller.rb
brucebolt: @issyl0 feel free to change this in your PR.  Mine was just a placeholder so I could add a test for the redirect.

PR #35, opened 2020-03-21
Add question 5.1 - What kind of space?

injms:
This has been mainly ~copied~ updated from 'What kind of medical equipment?' question, with the content updated, since the questions are similar in terms of the form fields and logic.

The logic for question 5 ('Can you offer space?') have been updated to point to this question if 'yes', and go to the next question if 'No'.

### DIFF
--- a/app/controllers/coronavirus_form/offer_space_controller.rb
+++ b/app/controllers/coronavirus_form/offer_space_controller.rb
@@ -22,6 +22,8 @@ class CoronavirusForm::OfferSpaceController < ApplicationController
       render "coronavirus_form/#{PAGE}"
     elsif session["check_answers_seen"]
       redirect_to controller: "coronavirus_form/check_answers", action: "show"
+    elsif offer_space.eql? "Yes"
+      redirect_to controller: "coronavirus_form/offer_space_type", action: "show"
     else
       redirect_to controller: "coronavirus_form/expert_advice", action: "show"
     end
--- a/app/views/coronavirus_form/offer_space_type.erb
+++ b/app/views/coronavirus_form/offer_space_type.erb
@@ -7,8 +7,6 @@
   <%= render "govuk_publishing_components/components/back_link", { href: previous_path } %>
 <% end %>
 
-<%= render "coronavirus_form/validation_error" %>
-
 <%= form_tag({},
   "data-module": "track-coronavirus-form-business-offer-space-type",
   "data-question-key": "what_kind_of_space",
@@ -17,7 +15,8 @@
   <%= render "govuk_publishing_components/components/checkboxes", {
     heading: t('coronavirus_form.offer_space_type.title'),
     is_page_heading: true,
-    name: "offer_space_type[]",
+    name: "offer_space_type",
+    error_message: error_items("offer_space_type[]"),
     items: [
       {
         value: t('coronavirus_form.offer_space_type.options.warehouse_space.label'),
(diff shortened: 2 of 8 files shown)


bilbof (commented):
Looks good! Please can we have a test for this one?

app/views/coronavirus_form/offer_space_type.erb
```diff
@@ -0,0 +1,47 @@
...
+<% content_for :meta_tags do %>
+  <meta name="description" content="<%= t('coronavirus_form.what_kind_of_space.title') %>" />
+<% end %>
+
+<% content_for :back_link do %>
+  <%= render "govuk_publishing_components/components/back_link", { href: previous_path } %>
+<% end %>
+
+<%= form_tag({},
+  "data-module": "track-coronavirus-form-business-offer-space",
```
richardTowers: Should this be different from the `offer_space` view?

app/controllers/coronavirus_form/offer_space_controller.rb
```diff
@@ -22,6 +22,8 @@ def submit
       render "coronavirus_form/#{PAGE}"
     elsif session["check_answers_seen"]
       redirect_to controller: "coronavirus_form/check_answers", action: "show"
+    elsif offer_space === "Yes"
```
huwd: I'm a little unsure if `===` or `==` is better pratice here.
I always get confused coming back from JS.

I know that `===` is implimented differently depending on the parent class.
But this seems to work here with the string matching, so feel free to ignore if this is pedentry!

spec/controllers/coronavirus_form/offer_space_controller_spec.rb
```diff
@@ -24,11 +24,16 @@
       expect(session[session_key]).to eq selected
     end
 
-    it "redirects to next step for a permitted response" do
-      post :submit, params: { offer_space: selected }
+    it "redirects to next step for a 'No' response" do
```
huwd, line 27: Much clearer, will update my tests to use that!

huwd (approved):

app/views/coronavirus_form/offer_space_type.erb
huwd: I think so, this gets rendered as a `data-attribute` in the HTML,
So we're using html attribute syntax wich is all kebabs i believe

injms: Yes, yes it should. Thank you!

app/controllers/coronavirus_form/offer_space_controller.rb
injms: Would `.eql?` be more Ruby?

injms: I've changed it to `.eql?` :)

spec/controllers/coronavirus_form/offer_space_type_controller_spec.rb
```diff
@@ -0,0 +1,75 @@
...
+
+    it "validates a valid option is chosen" do
+      post :submit, params: { offer_space: "<script></script>" }
+
+      expect(response).to render_template(current_template)
+    end
+
+    it "redirects to check your answers if check your answers previously seen" do
+      session[:check_answers_seen] = true
+      post :submit, params: { offer_space: selected }
```
bilbof: This should be `offer_space_type`

injms: Yes!

huwd (approved):

bilbof:
🚢 :shipit:

PR #36, opened 2020-03-21
Add question 9: Other kinds of support

huwd:
Trello: https://trello.com/c/AC88YKom/30-implement-question-9

# What
Adds qusetion 9, text area for other support.

# Screenshots
image

image

### DIFF
--- a/app/controllers/coronavirus_form/offer_other_support_controller.rb
+++ b/app/controllers/coronavirus_form/offer_other_support_controller.rb
@@ -1,0 +1,33 @@
+# frozen_string_literal: true
+
+class CoronavirusForm::OfferOtherSupportController < ApplicationController
+  include ActionView::Helpers::SanitizeHelper
+  include FieldValidationHelper
+
+  def show
+    render "coronavirus_form/#{PAGE}"
+  end
+
+  def submit
+    offer_other_support = sanitize(params[:offer_other_support]).presence
+
+    session[:offer_other_support] = offer_other_support
+
+    if session["check_answers_seen"]
+      redirect_to controller: "coronavirus_form/check_answers", action: "show"
+    elsif !session[:offer_other_support].nil?
+      redirect_to controller: "coronavirus_form/#{NEXT_PAGE}", action: "show"
+    else
+      redirect_to controller: "coronavirus_form/#{PAGE}", action: "show"
+    end
+  end
+
+private
+
+  PAGE = "offer_other_support"
+  NEXT_PAGE = "thank_you"
+
+  def previous_path
+    coronavirus_form_offer_community_support_path
+  end
+end
--- a/app/views/coronavirus_form/offer_other_support.html.erb
+++ b/app/views/coronavirus_form/offer_other_support.html.erb
@@ -1,0 +1,28 @@
+<% content_for :title do %><%= t('coronavirus_form.offer_other_support.title') %><% end %>
+<% content_for :meta_tags do %>
+  <meta name="description" content="<%= t('coronavirus_form.offer_other_support.title') %>" />
+<% end %>
+
+<% content_for :back_link do %>
+  <%= render "govuk_publishing_components/components/back_link", { href: previous_path } %>
+<% end %>
+
+<%= render "coronavirus_form/validation_error" %>
+<%= form_tag({},
+  "data-module": "track-coronavirus-form-offer-other-support",
+  "data-question-key": "offer_other_support",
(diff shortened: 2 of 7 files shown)


app/controllers/coronavirus_form/offer_other_support_controller.rb
```diff
@@ -0,0 +1,41 @@
...
+    end
+  end
+
+private
+
+  PAGE = "offer_other_support"
+  NEXT_PAGE = "your_business_details"
+
+  def previous_path
+    "/"
```
bilbof: Just need to update this bit

app/controllers/coronavirus_form/offer_other_support_controller.rb
```diff
@@ -0,0 +1,41 @@
...
+      PAGE,
+      other: support_other,
+    )
+
+    if invalid_fields.any?
+      flash.now[:validation] = invalid_fields
+      render "coronavirus_form/#{PAGE}"
+    elsif session["check_answers_seen"]
+      redirect_to controller: "coronavirus_form/check_answers", action: "show"
+    elsif session[:medical_equipment] == I18n.t("coronavirus_form.medical_equipment.options.option_yes.label")
```
bilbof: Does `medical_equipment` need to be `support_other` or could the line be removed?

huwd: yeah sorry, just put this up to have somewhere to park it, still [WIP]!

app/controllers/coronavirus_form/offer_other_support_controller.rb
```diff
@@ -0,0 +1,41 @@
...
+      PAGE,
+      other: support_other,
+    )
+
+    if invalid_fields.any?
+      flash.now[:validation] = invalid_fields
+      render "coronavirus_form/#{PAGE}"
+    elsif session["check_answers_seen"]
+      redirect_to controller: "coronavirus_form/check_answers", action: "show"
+    elsif session[:offer_other_support] == I18n.t("coronavirus_form.offer_other_support.options.option_yes.label")
```
bilbof: It looks like this line would need to be updated as we're missing an `option_yes.label` for the section.

bilbof (commented):
Just a minor comment.

Please could we have a test for this page?

I wonder if we could also set a character limit on the textarea (not just this one), since we're sticking this in a cookie for now which (iirc) has a 4kb limit.

huwd:
Yes sorry, should have been there to begin with!

app/controllers/coronavirus_form/offer_other_support_controller.rb
huwd: great added that, thanks for the spot

app/controllers/coronavirus_form/offer_other_support_controller.rb
```diff
@@ -0,0 +1,33 @@
...
+    end
+  end
+
+private
+
+  PAGE = "offer_other_support"
+  NEXT_PAGE = "thank_you"
+
+  def previous_path
+    "/"
```
bilbof:
```suggestion
    coronavirus_form_offer_community_support_path
```

bilbof (approved):
Minor change required, but otherwise good to go 👍 fab work

app/views/coronavirus_form/offer_other_support.html.erb
```diff
@@ -0,0 +1,29 @@
...
+<% end %>
+
+<%= render "coronavirus_form/validation_error" %>
+<%= form_tag({},
+  "data-module": "track-coronavirus-form-offer-other-support",
+  "data-question-key": "offer_other_support",
+  "id": "offer_other_support"
+) do %>
+
+<%# Terrible hack for now, will move this into the component properly before launch %>
```
alex-ju: Updating `govuk_publishing_component` to 21.32.0 will provide support for this use case. Example here: https://components.publishing.service.gov.uk/component-guide/textarea/with_label_as_page_heading.

app/controllers/coronavirus_form/offer_other_support_controller.rb
huwd: thanks very much, resolved!

app/views/coronavirus_form/offer_other_support.html.erb
huwd: thanks very much for that alex, have amended it!

PR #38, opened 2020-03-21
Add question 11: Contact details

bilbof:
This adds the contact details form, which is the final question before checking your answers.

https://trello.com/c/HZNOAAhW/36-implement-question-11-contact-details

### DIFF
--- a/app/views/coronavirus_form/contact_details.html.erb
+++ b/app/views/coronavirus_form/contact_details.html.erb
@@ -1,0 +1,55 @@
+<% content_for :title do %><%= t('coronavirus_form.contact_details.title') %><% end %>
+<% content_for :meta_tags do %>
+  <meta name="description" content="<%= t('coronavirus_form.contact_details.title') %>" />
+<% end %>
+
+<% content_for :back_link do %>
+  <%= render "govuk_publishing_components/components/back_link", { href: previous_path } %>
+<% end %>
+
+<%= render "govuk_publishing_components/components/title", {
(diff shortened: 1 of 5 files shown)


app/views/coronavirus_form/contact_details.html.erb
```diff
@@ -0,0 +1,55 @@
...
+  title:t("coronavirus_form.contact_details.title"),
+  margin_top: 0
+} %>
+
+<%= form_tag({},
+  "data-module": "track-coronavirus-form-business-contact-details",
+  "data-question-key": "contact_details",
+  "id": "contact_details"
+) do %>
+<%= render "govuk_publishing_components/components/input", {
```
huwd, line 20: do we indent this, given the `do`?

config/routes.rb
```diff
@@ -53,6 +53,14 @@
   get "/coronavirus-form/offer-community-support" => "coronavirus_form/offer_community_support#show"
   post "/coronavirus-form/offer-community-support" => "coronavirus_form/offer_community_support#submit"
 
+  # Question 11
```
huwd, line 56: Great idea! Will add these back in above other routes too!

app/views/coronavirus_form/contact_details.html.erb
bilbof: We haven't elsewhere, so I left it like that for consistency.

huwd (approved):

PR #41, opened 2020-03-21
Add submit method to check your answers

theseanything:
- This redirects users to the thank you page
- Saves form response to the database
- Clears the session

### DIFF
--- a/app/controllers/coronavirus_form/check_answers_controller.rb
+++ b/app/controllers/coronavirus_form/check_answers_controller.rb
@@ -7,4 +7,23 @@ class CoronavirusForm::CheckAnswersController < ApplicationController
     session[:check_answers_seen] = true
     render "coronavirus_form/check_answers"
   end
+
+  def submit
+    submission_reference = reference_number
+
+    session[:reference_number] = submission_reference
+    # FormResponse.create(form_response: session)
+
+    reset_session
+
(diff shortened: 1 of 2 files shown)


theseanything:
Needs error handling, but could be subsequent PR

app/controllers/coronavirus_form/check_answers_controller.rb
```diff
@@ -7,4 +7,23 @@ def show
     session[:check_answers_seen] = true
     render "coronavirus_form/check_answers"
   end
+
+  def submit
+    submission_reference = reference_number
+
+    session[:reference_number] = submission_reference
+    FormResponse.create(form_response: session)
```
bilbof, line 15: I guess this is still a work in progress, but it would be cool to get some validations on the data we're accepting ( this is probably more fiddly with a jsonb format ), like text length etc.

theseanything: Yeah was thinking about this, do we do validation in the individual questions and here? I guess we should also revalidate here as someone could modify the cookie or hit this endpoint directly

theseanything: Could do in a separate PR?

bilbof: OK sounds good. I'd like validation here if possible since we might mess up/forget validation elsewhere.

bilbof (approved):

PR #42, opened 2020-03-21
Add pry and awesome print

huwd:
# What

Switch to using pry,
Also updates rails to 6.0.2.2 as that was a dependency of the latest pry.

# Why

I generally find pry easier to work with when dropping in a debugger, pry rails makes pry the default for `rails c` and awesome_print gives you nicely coloured and formatted outputs for things like hashes.

Feel free to shoot this down if the majority's workflow expects IRB?


bilbof (approved):
Thanks Huw!

PR #43, opened 2020-03-21
Add question 10: business details

brucebolt:
Adds question 10 to collect business details.  Implements the relevant routes, validation, controller and view, plus a placeholder route to the following question.

Screenshot 2020-03-21 at 12 05 22

Trello card: https://trello.com/c/7DoCeLfM

### DIFF
--- a/app/controllers/coronavirus_form/business_details_controller.rb
+++ b/app/controllers/coronavirus_form/business_details_controller.rb
@@ -1,0 +1,54 @@
+# frozen_string_literal: true
+
+class CoronavirusForm::BusinessDetailsController < ApplicationController
+  include ActionView::Helpers::SanitizeHelper
+  include FieldValidationHelper
+  include FormFlowHelper
+
+  before_action :check_first_question_answered, only: :show
+
+  REQUIRED_FIELDS = %w(company_name).freeze
+
+  def show
+    render "coronavirus_form/#{PAGE}"
+  end
+
+  def submit
+    @business_details = sanitized_business_details(params)
+    session[:business_details] = @business_details
+
+    invalid_fields = validate_fields(@business_details)
+
+    if invalid_fields.any?
+      flash.now[:validation] = invalid_fields
+      render "coronavirus_form/#{PAGE}"
+    else
+      redirect_to controller: session["check_answers_seen"] ? "coronavirus_form/check_answers" : "coronavirus_form/#{NEXT_PAGE}", action: "show"
+    end
+  end
+
+private
+
+  NEXT_PAGE = "contact_details"
+  PAGE = "business_details"
+
+  def validate_fields(business_details)
+    missing_fields = validate_mandatory_text_fields(REQUIRED_FIELDS, PAGE)
+    company_size_validation = validate_radio_field("#{PAGE}.company_size", radio: business_details["company_size"])
+    company_location_validation = validate_radio_field("#{PAGE}.company_location", radio: business_details["company_location"])
+    missing_fields + company_size_validation + company_location_validation
+  end
+
+  def sanitized_business_details(params)
+    {
+      "company_name" => sanitize(params[:company_name]).presence,
+      "company_number" => sanitize(params[:company_number]).presence,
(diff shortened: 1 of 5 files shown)


config/routes.rb
```diff
@@ -65,6 +65,9 @@
   post "/coronavirus-form/business-details" => "coronavirus_form/business_details#submit"
 
   # Question 12
+  get "/coronavirus-form/business-details" => "coronavirus_form/business_details#show"
```
erino: This route and the one below look to be the same as those on line 64 and 65?

brucebolt: Good spot!  Came from a rebase that didn't quite go to plan.

app/controllers/coronavirus_form/business_details_controller.rb
```diff
@@ -0,0 +1,64 @@
...
+  PAGE = "business_details"
+
+  def validate_fields(business_details)
+    missing_fields = validate_missing_fields(business_details)
+    company_size_validation = validate_radio_field("#{PAGE}.company_size", radio: business_details["company_size"])
+    company_location_validation = validate_radio_field("#{PAGE}.company_location", radio: business_details["company_location"])
+    missing_fields + company_size_validation + company_location_validation
+  end
+
+  def validate_missing_fields(business_details)
```
bevanloon: is it worth using the helper https://github.com/alphagov/govuk-coronavirus-business-volunteer-form/blob/master/app/helpers/field_validation_helper.rb#L4 instead?

app/controllers/coronavirus_form/business_details_controller.rb
```diff
@@ -0,0 +1,64 @@
...
+    end
+  end
+
+  def sanitized_business_details(params)
+    {
+      "company_name" => sanitize(params[:company_name]).presence,
+      "company_number" => sanitize(params[:company_number]).presence,
+      "company_size" => sanitize(params[:company_size]).presence,
+      "company_location" => sanitize(params[:company_location]).presence,
+    }
```
bilbof, line 48: It would be nice to validate later that the options the user has chosen are the ones we've defined.

spec/controllers/coronavirus_form/business_details_controller_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+    end
+
+    it "validates company size option is chosen" do
+      post :submit, params: { company_size: [] }
+
+      expect(response).to render_template(current_template)
+    end
+
+    it "validates company location option is chosen" do
+      post :submit, params: { company_location: [] }
```
bilbof: It looks like `company_size` and `company_location` should be `nil` or `""` since radio button values get submitted as strings.

bilbof (approved):
Looks good; minor comments 👍

app/views/coronavirus_form/business_details.html.erb
```diff
@@ -0,0 +1,65 @@
...
+<% end %>
+
+<%= form_tag({},
+  "data-module": "track-coronavirus-form-business-business-details",
+  "data-question-key": "business_details",
+  "id": "business_details"
+) do %>
+
+<%= render "govuk_publishing_components/components/title", {
+  title: t('coronavirus_form.business_details.title'),
```
alex-ju, line 17: This needs a `margin_top: 0`

app/controllers/coronavirus_form/business_details_controller.rb
brucebolt: Good spot.  We're not actually using `validate_mandatory_text_fields` anywhere 🤔 .

app/controllers/coronavirus_form/business_details_controller.rb
brucebolt: Maybe this is something we should add to `validate_radio_field` post-release?

bilbof: sounds good 👍

bilbof (approved):

PR #44, opened 2020-03-21
Update existing content and validation messages

alex-ju:
Update existing content and validation messages.

[Trello card](https://trello.com/c/yCotfI0T)

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -1,7 +1,24 @@
 en:
   start_page:
-    title: Offer support as a business to help with coronavirus
-    body: Use this service to share the types of things your business can help with or might be able to help with. You'll be contacted with more information about how you can help.
+    title: Offer coronavirus support from your business
+    body: |
+      <p>Use this service to share the types of support you think your business can offer.</p>
+      <p>The support needed includes:</p>
+      <ul>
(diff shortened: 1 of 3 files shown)


bilbof (approved):
Nice one!

config/locales/en.yml
```diff
@@ -1,7 +1,24 @@
 en:
   start_page:
-    title: Offer support as a business to help with coronavirus
-    body: Use this service to share the types of things your business can help with or might be able to help with. You'll be contacted with more information about how you can help.
+    title: Offer coronavirus support from your business
+    body: |
+      <p>Use this service to share the types of support you think your business can offer.</p>
```
injms, line 5: Worth using `govuk-body` and `govuk-list govuk-list--bullet`?

alex-ju: Ah, sorry, merged too quickly. I passed it through govspeak as I think it's easier for content designers to edit yml files without having to worry they're breaking things (such as mistakenly removing a double quote when editing content). Arguably missing a closing bracket might be similar, but browsers are often forgiving with these.

alex-ju: Ideally, this would've been converted from markdown straight away, without having to add any tags, but we don't have a helper for that atm.

PR #45, opened 2020-03-21
Run migrations in Procfile

richardTowers:
This will make the migrations happen in PaaS. Note we're not using
db:setup here because the DB already exists in PaaS, so we don't need to
create it. Also we've removed the db name from the config because DB
names are auto-generated by PaaS.

### DIFF
--- a/.cfignore
+++ b/.cfignore
@@ -1,3 +1,2 @@
 tmp
-Procfile
 vendor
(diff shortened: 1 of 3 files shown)


theseanything (approved):

alex-ju:
@richardTowers I've added a commit to fix linting in Gemfile so we can get this merged, hope you don't mind.

update: now that you addressed it separately this should go away at rebase/merge

richardTowers:
Thanks @alex-ju! Rebased.

PR #55, opened 2020-03-23
Remove /coronavirus-form from all routes

brucebolt:
Form questions will now be served from the root.  This matches https://github.com/alphagov/govuk-coronavirus-vulnerable-people-form/pull/101.


issyl0 (approved):
Woohoo!

PR #57, opened 2020-03-23
Update "Can you offer expert advice?" question

brucebolt:
Changes the question to a 'Yes'/'No' response.  If user selects 'Yes', they are taken to an additional question asking specifically the types of advice they can offer.

Screenshot 2020-03-23 at 15 06 42
Screenshot 2020-03-23 at 15 10 34

Trello card: https://trello.com/c/HCXOxjTm

### DIFF
--- a/app/controllers/coronavirus_form/expert_advice_controller.rb
+++ b/app/controllers/coronavirus_form/expert_advice_controller.rb
@@ -5,46 +5,35 @@ class CoronavirusForm::ExpertAdviceController < ApplicationController
   include FieldValidationHelper
 
   def show
-    session[:expert_advice] ||= []
     render "coronavirus_form/#{PAGE}"
   end
 
   def submit
-    expert_advice = Array(params[:expert_advice]).map { |item| sanitize(item).presence }.compact
-    expert_advice_other = sanitize(params[:expert_advice_other]).presence
+    expert_advice = sanitize(params[:expert_advice]).presence
     session[:expert_advice] = expert_advice
-    session[:expert_advice_other] = if selected_other?(expert_advice)
-                                      expert_advice_other
-                                    else
-                                      ""
-                                    end
-    invalid_fields = validate_checkbox_field(
+
+    invalid_fields = validate_radio_field(
       PAGE,
-      values: expert_advice,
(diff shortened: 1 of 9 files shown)


app/controllers/coronavirus_form/expert_advice_type_controller.rb
```diff
@@ -0,0 +1,50 @@
+# frozen_string_literal: true
+
+class CoronavirusForm::ExpertAdviceTypeController < ApplicationController
+  include ActionView::Helpers::SanitizeHelper
+  include FieldValidationHelper
+
+  def show
+    session[:expert_advice] ||= []
```
bilbof: We're re-using the session key `expert_advice` so this should be e.g. `expert_advice_type`.

app/views/coronavirus_form/expert_advice_type.html.erb
```diff
@@ -0,0 +1,68 @@
...
+    <%= form_tag({},
+      "data-module": "track-coronavirus-form-business-medical-equipment-type",
+      "data-question-key": "expert_advice",
+      "id": "expert_advice_type"
+    ) do %>
+    <%= render "govuk_publishing_components/components/checkboxes", {
+      heading: t("coronavirus_form.expert_advice_type.title"),
+      hint_text: t("coronavirus_form.expert_advice_type.hint"),
+      is_page_heading: true,
+      name: "expert_advice[]",
```
bilbof: This should be changed to `expert_advice_type` too

bilbof (changes requested):
Looks good, just a couple of small things, and the tests are failing.

app/controllers/coronavirus_form/expert_advice_controller.rb
```diff
@@ -5,46 +5,35 @@ class CoronavirusForm::ExpertAdviceController < ApplicationController
   include FieldValidationHelper
 
   def show
-    session[:expert_advice] ||= []
```
huwd, line 8: Is there not an advantage in memosing something here?

brucebolt: We're now storing a Yes/No value instead of an array, so there's no need to set a default value anymore

bilbof (approved):

PR #59, opened 2020-03-23
Redirect user to first question if they have not answered it

brucebolt:
This will force users who come in through the URL of an intermediate question to return to the start of the form.

Trello card: https://trello.com/c/ZsjDHDv2

### DIFF
--- a/app/helpers/form_flow_helper.rb
+++ b/app/helpers/form_flow_helper.rb
@@ -1,0 +1,13 @@
+module FormFlowHelper
+  FIRST_QUESTION = "medical_equipment".freeze
+
+  def check_first_question_answered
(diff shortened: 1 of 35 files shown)


bilbof (approved):

app/helpers/form_flow_helper.rb
```diff
@@ -0,0 +1,13 @@
...
+  FIRST_QUESTION = "medical_equipment".freeze
+
+  def check_first_question_answered
+    redirect_to controller: "coronavirus_form/#{FIRST_QUESTION}", action: "show" unless first_question_answered?
+  end
+
+private
+
+  def first_question_answered?
+    session["medical_equipment"].present?
```
huwd:
```suggestion
    session["#{FIRST_QUESTION}"].present?
```

huwd (approved):

PR #63, opened 2020-03-23
Add care qualifications question

brucebolt:
Adds a new question which is presented only if user selects "Yes" to "Can you offer social care or childcare?".

Trello card: https://trello.com/c/lLpZPqJN


bilbof (commented):
Looks good 🎉 

It might be good to have an error message on the NHS textarea, since it's required, and it's not obvious the text box is required if the box isn't filled in. I'm not sure why what you have didn't work.

Screenshot 2020-03-23 at 19 39 45

bilbof (approved):

PR #64, opened 2020-03-23
Question 1.3 updates (product page)

bilbof:
Two updates to the product page. This page is a little complicated; there are details in the commits. 

https://trello.com/c/BlREN0tI/99-update-13-tell-us-about-the-product-youre-offering-page

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -71,16 +71,27 @@ en:
       custom_select_error: Select the best way to describe your business
     product_details:
       title: Tell us about each product you’re offering
+      hint: Give details for one product at a time.
       product_name:
         label: Name of product
       product_cost:
-        label: Cost of product
-        hint: For example, £23.99. Enter 0 if you will donate it.
+        label: Cost of product in pounds
+        hint: For example, 23.99. Enter 0 if you will donate it.
       certification_details:
         label: Certification details
         hint: For example, CE marking or EN standards.
-      product_postcode:
-        label: Postcode of stock location or manufacture
+      product_location:
+        label: Where is the product made, or stored if it's already made?
+        options:
+          option_uk:
+            label: United Kingdom
+            input:
+              label: Postcode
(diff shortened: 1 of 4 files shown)


config/locales/en.yml
```diff
@@ -72,20 +72,52 @@ en:
       custom_select_error: Select the best way to describe your business
     product_details:
       title: Tell us about each product you're offering
```
alex-ju:
```suggestion
      title: Tell us about the product you’re offering
```

config/locales/en.yml
```diff
@@ -72,20 +72,52 @@ en:
...
+      hint: Give details for one product at a time.
```
alex-ju, line 74: We don't need this one here it seems

alex-ju (commented):
🎉 I left two minor content comments. Frontend looks good and it works as expected, but I would like Bruce and/or Huw to have a look tomorrow morning.

config/locales/en.yml
bilbof: I'm going to do this in another PR, since there are other updates to this bit: https://trello.com/c/koG4jHHw/113-change-format-and-flow-of-question-1

alex-ju: 👍  I was just looking at that card

config/locales/en.yml
bilbof: This is included in the new designs (see 1.1)

alex-ju (approved):
It looks great, I'm confident we can get it merged and update later on if needed 🥇

config/locales/en.yml
alex-ju: I was looking at 1.3

bilbof:
Thanks @alex-ju 👍 ✋

config/locales/en.yml
bilbof: Ah sorry, I'll fix any issues in a followup PR

PR #66, opened 2020-03-23
Update content and tidy up existing views

alex-ju:
A few changes to align the current views with the prototype:
- Update content on start page
- Use apostrophes
- Update existing content across service
- Add missing hint text
- Use small headings on business details
- Remove redundant validation error


bilbof (approved):
🎉

PR #69, opened 2020-03-24
Fix validation error message for contact details form

brucebolt:
The word "name" appears to be reserved for the error messages.  Currently, an invalid email address will also show alongside the "Name of main contact field".  Therefore changing the name of this field to something that isn't reserved.

Example:
Screenshot 2020-03-24 at 08 27 32


bilbof (approved):
Nice one!

PR #98, opened 2020-03-24
Add redis as session store

bilbof:
This allows users to input over 4kb of data.

https://trello.com/c/CdDdV2fc/120-use-redis-as-session-store

Uses https://github.com/redis-store/redis-rails

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -4,13 +4,13 @@ ruby File.read(".ruby-version").strip
 
 source "https://rubygems.org"
 
-gem "rails", "~> 6.0.2"
-
 gem "bootsnap", "~> 1"
 gem "dynamoid"
 gem "govuk_app_config", "~> 2.1.1"
 gem "govuk_publishing_components", "~> 21.34.1"
 gem "puma", "~> 4.3"
+gem "rails", "~> 6.0.2"
+gem "redis"
 gem "sass-rails", "< 6"
 gem "sentry-raven", "~> 3.0"
 gem "uglifier", "~> 4.2"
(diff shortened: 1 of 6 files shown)


theseanything:
How do we set the Redis URL etc?

README.md
```diff
@@ -35,6 +35,13 @@ Then set up your Docker database
 
 You'll then need to specify the `DATABASE_URL` environment variable before the below tasks.
 
+### Running Redis
+
+#### Locally
+
+    brew install redis
+    redis-server
```
theseanything: Nit: but it think you can use `brew services start redis`

Gemfile
```diff
@@ -4,13 +4,15 @@ ruby File.read(".ruby-version").strip
...
-gem "rails", "~> 6.0.2"
 
 gem "bootsnap", "~> 1"
 gem "dynamoid"
 gem "govuk_app_config", "~> 2.1.1"
 gem "govuk_publishing_components", "~> 21.34.0"
 gem "puma", "~> 4.3"
+gem "rails", "~> 6.0.2"
+gem "redis"
+gem "redis-rails", "~> 5.0"
```
theseanything: Do we need this? Seems to be in-built - https://github.com/redis-store/redis-rails#a-quick-note-about-rails-52?

bilbof: I wonder if that also applies to sessions? I'm using that gem for the session stuff:

> We are still actively maintaining all other gems in the redis-store family, such as redis-actionpack for session management

huwd:
Humm, why did the deployment pipeline here try and go to heroku?
`coronavirus-redis-sessi-bwz0hn`?

huwd (approved):
@bilbof might need to bundle install and push?
Conflict in the lockfile.

Otherwise LGTM, if `redis-rails` is still a thing!

Gemfile
theseanything: Yeah, it's not clear from the documentation. I guess if this works, we can look at it later, if its an issue!

theseanything (approved):

rjbaker (approved):
Looks good.

PR #106, opened 2020-03-25
Limit all text fields to 1000 characters

brucebolt:
Ensures that all text fields are 1000 or fewer characters

Trello card: https://trello.com/c/8kLHw5Hq

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,6 +1,12 @@
 # frozen_string_literal: true
 
 class ApplicationController < ActionController::Base
+  include ActionView::Helpers::SanitizeHelper
+  include CheckAnswersHelper
+  include FieldValidationHelper
+  include FormFlowHelper
+  include ProductHelper
+
   if ENV["REQUIRE_BASIC_AUTH"]
     http_basic_authenticate_with(
       name: ENV.fetch("BASIC_AUTH_USERNAME"),
--- a/app/controllers/coronavirus_form/contact_details_controller.rb
+++ b/app/controllers/coronavirus_form/contact_details_controller.rb
@@ -1,10 +1,6 @@
 # frozen_string_literal: true
 
 class CoronavirusForm::ContactDetailsController < ApplicationController
-  include ActionView::Helpers::SanitizeHelper
-  include FieldValidationHelper
-  include FormFlowHelper
-
   before_action :check_first_question_answered, only: :show
 
   REQUIRED_FIELDS = %w(contact_name phone_number).freeze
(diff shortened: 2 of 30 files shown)


app/controllers/application_controller.rb
```diff
@@ -1,6 +1,12 @@
 # frozen_string_literal: true
 
 class ApplicationController < ActionController::Base
+  include ActionView::Helpers::SanitizeHelper
```
huwd, line 4: nice refactor

app/controllers/coronavirus_form/contact_details_controller.rb
```diff
@@ -1,13 +1,10 @@
...
 
 class CoronavirusForm::ContactDetailsController < ApplicationController
-  include ActionView::Helpers::SanitizeHelper
-  include FieldValidationHelper
-  include FormFlowHelper
-
   before_action :check_first_question_answered, only: :show
 
   REQUIRED_FIELDS = %w(contact_name phone_number).freeze
+  TEXT_FIELDS = %w(contact_name role phone_number email).freeze
```
huwd, line 7: Watch out for symbol vs string clashes - this caught us out a bit in the vulnerable people form?

brucebolt: Yeah, this doesn't seem to have been a problem here

app/helpers/field_validation_helper.rb
```diff
@@ -1,6 +1,19 @@
 # frozen_string_literal: true
 
 module FieldValidationHelper
+  def validate_field_response_length(page, fields)
+    invalid_fields = []
+    fields.each do |field|
+      next unless params[field]
```
huwd, line 7: so... params are not in the session and are always string?
Field is from that word array in each controller... so these match?

config/locales/en.yml
```diff
@@ -176,6 +179,7 @@ en:
         custom_select_error: Select if you can offer another product
       offer_space_type:
         title: What kind of space can you offer?
+        field: something
```
huwd: what is this key?

brucebolt: Oops!

app/controllers/coronavirus_form/contact_details_controller.rb
huwd: I think this checks out because you're using the params:
https://github.com/alphagov/govuk-coronavirus-business-volunteer-form/pull/106#discussion_r397735176

app/helpers/field_validation_helper.rb
brucebolt: The params hash has `HashWithIndifferentAccess` enabled, so we can use either strings or symbols to access

huwd (approved):

huwd:
Looks great, nice one!

PR #108, opened 2020-03-25
Update previous links to match design

brucebolt:
Some previous paths were incorrect, resulting in the Back button giving an incorrect user journey.

The flow now matches the design: https://www.figma.com/file/yjWy0c1LxyxB237K5IjqAh/Business-volunteering?node-id=0%3A1

Trello card: https://trello.com/c/3EOVtgeV


issyl0 (approved):
Cross-referenced with Figma and it looks right to me.

PR #109, opened 2020-03-25
Change Dynimoid logging level to :warn

huwd:
[Trello Card](https://trello.com/c/99xm7mlq/86-ensure-sensitive-data-doesnt-go-to-logit-splunk)

# What
[Dynimod config.logger is a rails logger by default](https://github.com/Dynamoid/Dynamoid#configuration), rails logging comes with `:debug`, `:info`, `:warn`, `:error`, `:fatal`, and `:unknown`.

We were saying `:info` was TMI, but `:error` is perhaps overly cautious.
Lets see what `:warn` gets us

# Why
Default logging behaviour with Dynimod is to log everything, so we need to calm it down it bit to keep our logs free of PII.

### DIFF
--- a/config/initializers/dynamoid.rb
+++ b/config/initializers/dynamoid.rb
@@ -8,7 +8,7 @@ Dynamoid.configure do |config|
   config.access_key = ENV["AWS_ACCESS_KEY_ID"]
   config.secret_key = ENV["AWS_SECRET_ACCESS_KEY"]
   config.region = "eu-west-2"
-  config.logger.level = :error
+  config.logger.level = :warn
 
   unless Rails.env.production?
     # [Optional]. If provided, it communicates with the DB listening at the endpoint - to communication with DynamoDB local.


huwd:
@richardTowers is there any way to test this against splunk / loggit without deploying to staging?
Could see the dev logs otherwise?

config/initializers/dynamoid.rb
```diff
@@ -8,7 +8,7 @@
   config.access_key = ENV["AWS_ACCESS_KEY_ID"]
   config.secret_key = ENV["AWS_SECRET_ACCESS_KEY"]
   config.region = "eu-west-2"
-  config.logger.level = :error
+  config.logger.level = :warn
```
richardTowers, line 11: I think the problem here is that `config.logger` is a reference to the global rails logger.

So this line is setting the log level _everywhere_. At the moment, validation errors are logged at info, so you still won't get them with this.

richardTowers:
> @richardTowers is there any way to test this against splunk / loggit without deploying to staging?
> Could see the dev logs otherwise?

Yep! `cf logs $appname`.

Although you'd still need to get it deployed somewhere 🤔 

You could try getting it to work with RAILS_ENV=production locally. Bit of a yak shave perhaps.

config/initializers/dynamoid.rb
huwd: :thinking: humm, ok back to the drawing board on that one then.
Currently it's set to error same as the vulnerable people form.

I think i need to be a bit clearer on the state we want (sounds like, no PII but getting validation errors in logs), and where we should/souldn't see those.

Can catch up when we've got the more important stuff out of the way?

richardTowers: Sure thing!

PR #113, opened 2020-03-25
Reorder yaml declarations to match journey

alex-ju:
This is required so that check your answers page has the same order as the journey.

[Trello card](https://trello.com/c/IAXffAQ4)


bilbof (approved):
Thank you 🙏

issyl0 (approved):

PR #115, opened 2020-03-25
Add novalidate to all forms

bilbof:
https://trello.com/c/0VrU0lXp/136-add-novalidate


issyl0 (approved):
I'd like to know the reasons why this has to be done. Maybe it's in the Trello ticket that I've not read. 🙈

huwd:
It prevents HTML form validation, I think we want all validation being run serverside

bilbof:
@issyl0 this was to prevent HTML5 behaviours we didn't want I believe. E.g. email validation on the client side.

PR #121, opened 2020-03-25
Force SSL for Secure cookies

bilbof:


huwd (approved):
Sounds very useful

bilbof:
I'm not sure what effect this will have on pre-existing cookies. I'm a little worried that it will invalidate pre-existing sessions. I'll merge it and see what happens to my cookie on staging.

PR #128, opened 2020-03-26
Disable basic auth in Production

issyl0:
- Maybe this'll work?


issyl0:
Wait no hold on, I remember someone saying something about how for the vulnerable people form it looks for the presence of the envvar not its value. So maybe blank was good? But why would it be re-enabled now?

issyl0:
Fixed. We think.

brucebolt (approved):

PR #129, opened 2020-03-26
Remove "Can you offer community support?" question

issyl0:
After "Can you offer social or child care?", the next question is *now*
"Can you offer any other kind of support?"

Can someone please check that I've routed things correctly for previous
and next?

https://trello.com/c/88EdCXUF/148-remove-can-you-offer-community-support-screen-8


app/controllers/coronavirus_form/offer_other_support_controller.rb
```diff
@@ -32,6 +32,6 @@ def submit
   NEXT_PAGE = "business_details"
 
   def previous_path
-    offer_community_support_path
+    offer_care_qualifications_path
```
brucebolt:
```suggestion
    offer_care_path
```
This should be the offer care page, since we don't know whether they answered Yes/No to whether they can offer care, so may not have seen the qualification page.

brucebolt (commented):
Looks good.  I've just left one comment about a previous path.

brucebolt (approved):

PR #133, opened 2020-03-26
Fix additional product index

alex-ju:
Make sure the "Add additional product button" is correctly placed and the summary list is plit at the right point (after products).

Questions index is not the same with items index (as the journey is not linear and people can add multiple products).

Example below with no products:

<table>
<tr><th>Before</th><th>After</th></tr>
<tr><td valign="top">

Screenshot 2020-03-26 at 16 59 02

</td><td valign="top">

Screenshot 2020-03-26 at 16 58 34

</td></tr>
</table>

### DIFF
--- a/app/controllers/coronavirus_form/check_answers_controller.rb
+++ b/app/controllers/coronavirus_form/check_answers_controller.rb
@@ -66,12 +66,16 @@ private
     end
   end
 
+  def additional_product_index
+    items.index { |item| item[:field] === t("coronavirus_form.questions.additional_product.title") }
+  end
+
   def items_part_1
-    items.select.with_index { |_, index| index < questions.index("additional_product") }
+    items.select.with_index { |_, index| index < additional_product_index }
   end
 
   def items_part_2
-    items.select.with_index { |_, index| index > questions.index("additional_product") }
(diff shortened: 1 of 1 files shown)


app/controllers/coronavirus_form/check_answers_controller.rb
```diff
@@ -66,12 +66,18 @@ def items
     end
   end
 
+  def additional_product_index
+    product_index = nil
+    items.each.with_index { |item, index| product_index = index if item[:field] === t("coronavirus_form.questions.additional_product.title") }
```
alex-ju: Not my best piece of work for sure, but don't know how to do it nicer in Ruby.

brucebolt: I think you might be able to do something like:

```
items.index { |item| item[:field] === t("coronavirus_form.questions.additional_product.title") }
```

Then you wouldn't need the assignment or explicit return.

alex-ju: This looks way better! Thank you 🙌

brucebolt (approved):
👍

PR #145, opened 2020-03-27
Use fragment caching for views

bilbof:
This is an attempt to improve on an already quite fast response time; feel free to close if this is unnecessary / confusing, since I'm not sure if we'll see big gains from this at our current scale.

Since we can't cache pages based on the URL (due to sessions), we can instead make use of [Rails 'fragment' caching](https://guides.rubyonrails.org/caching_with_rails.html#fragment-caching).

We can cache 'fragments' of the views. If two users view the same thing, e.g. a page without any user input, we can serve a fragment from the cache.

### DIFF
--- a/app/controllers/coronavirus_form/additional_product_controller.rb
+++ b/app/controllers/coronavirus_form/additional_product_controller.rb
@@ -18,7 +18,7 @@ class CoronavirusForm::AdditionalProductController < ApplicationController
     if invalid_fields.any?
       flash.now[:validation] = invalid_fields
       render "coronavirus_form/#{PAGE}"
-    elsif additional_product == I18n.t("coronavirus_form.questions.additional_product.options.option_yes.label")
+    elsif additional_product == YES
       redirect_to controller: "coronavirus_form/medical_equipment_type", action: "show"
     elsif session["check_answers_seen"]
       redirect_to controller: "coronavirus_form/check_answers", action: "show"
@@ -30,6 +30,7 @@ class CoronavirusForm::AdditionalProductController < ApplicationController
 private
 
   PAGE = "additional_product"
+  YES = I18n.t("coronavirus_form.questions.additional_product.options.option_yes.label")
 
   def previous_path
     session["product_details"] ||= []
--- a/app/controllers/coronavirus_form/check_answers_controller.rb
+++ b/app/controllers/coronavirus_form/check_answers_controller.rb
@@ -10,24 +10,26 @@ class CoronavirusForm::CheckAnswersController < ApplicationController
   end
 
   def submit
-    submission_reference = reference_number
-
-    session[:reference_number] = submission_reference
+    session[:reference_number] = reference_number
     FormResponse.create(form_response: session)
 
     reset_session
 
-    redirect_to controller: "coronavirus_form/thank_you", action: "show", reference_number: submission_reference
+    redirect_to controller: "coronavirus_form/thank_you", action: "show", reference_number: reference_number
   end
 
 private
 
   helper_method :items_part_1, :items_part_2
 
+  QUESTIONS = YAML.load_file(Rails.root.join("config/locales/en.yml")).dig("en", "coronavirus_form", "questions").keys
+
   def reference_number
-    timestamp = Time.zone.now.strftime("%Y%m%d-%H%M%S")
-    random_id = SecureRandom.hex(3).upcase
-    "#{timestamp}-#{random_id}"
+    @reference_number ||= begin
+      timestamp = Time.zone.now.strftime("%Y%m%d-%H%M%S")
+      random_id = SecureRandom.hex(3).upcase
+      "#{timestamp}-#{random_id}"
+    end
   end
 
   def previous_path
@@ -71,15 +73,17 @@ private
   end
 
   def additional_product_index
-    items.index { |item| item[:field] === t("coronavirus_form.questions.additional_product.title") }
+    @additional_product_index ||= items.index do |item|
+      item[:field] == t("coronavirus_form.questions.additional_product.title")
+    end
   end
 
   def items_part_1
-    items.select.with_index { |_, index| index < additional_product_index }
+    items.slice(0, additional_product_index)
   end
 
   def items_part_2
-    items.select.with_index { |_, index| index > additional_product_index }
+    items.slice(additional_product_index + 1..)
   end
 
   def product_details(products)
@@ -160,7 +164,6 @@ private
   end
 
   def questions
-    @questions ||= YAML.load_file(Rails.root.join("config/locales/en.yml"))
-    @questions["en"]["coronavirus_form"]["questions"].keys
+    @questions ||= QUESTIONS
   end
 end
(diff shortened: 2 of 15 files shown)


app/controllers/coronavirus_form/additional_product_controller.rb
```diff
@@ -18,7 +18,7 @@ def submit
     if invalid_fields.any?
       flash.now[:validation] = invalid_fields
       render "coronavirus_form/#{PAGE}"
-    elsif additional_product == I18n.t("coronavirus_form.questions.additional_product.options.option_yes.label")
+    elsif additional_product == YES
```
huwd, line 21: Was this just a nice tidy whilst you were here, or does this help with fragment caching too?

app/controllers/coronavirus_form/check_answers_controller.rb
```diff
@@ -10,24 +10,26 @@ def show
...
 
   helper_method :items_part_1, :items_part_2
 
+  QUESTIONS = YAML.load_file(Rails.root.join("config/locales/en.yml")).dig("en", "coronavirus_form", "questions").keys
+
   def reference_number
-    timestamp = Time.zone.now.strftime("%Y%m%d-%H%M%S")
-    random_id = SecureRandom.hex(3).upcase
-    "#{timestamp}-#{random_id}"
+    @reference_number ||= begin
```
huwd, line 28: ~~So fragment caching == memoizing any variables that have computational overhead?~~

Ah ha, [i see so fragment caching is a view layer thing](https://github.com/alphagov/govuk-coronavirus-business-volunteer-form/pull/145/files#r400184786). So you're just also memoizing here for another boost?

app/controllers/coronavirus_form/check_answers_controller.rb
```diff
@@ -67,15 +69,17 @@ def items
...
   def additional_product_index
-    items.index { |item| item[:field] === t("coronavirus_form.questions.additional_product.title") }
+    @additional_product_index ||= items.index do |item|
+      item[:field] == t("coronavirus_form.questions.additional_product.title")
+    end
   end
 
   def items_part_1
-    items.select.with_index { |_, index| index < additional_product_index }
+    items.slice(0, additional_product_index)
```
huwd, line 82: This is really nice, and much easier to grok!

app/controllers/coronavirus_form/check_answers_controller.rb
```diff
@@ -67,15 +69,17 @@ def items
...
   end
 
   def items_part_2
-    items.select.with_index { |_, index| index > additional_product_index }
+    items.slice(additional_product_index + 1..)
```
huwd, line 86: ditto here, really nice. So the 1.. == from the 2nd value to the end of the range right?

app/controllers/coronavirus_form/expert_advice_type_controller.rb
```diff
@@ -38,11 +38,11 @@ def submit
 
   PAGE = "expert_advice_type"
   NEXT_PAGE = "offer_care"
+  ALLOWED_VALUES = I18n.t("coronavirus_form.questions.#{PAGE}.options").map { |_, item| item.dig(:label) }
```
huwd, line 41: what would stop you memoizing this in a similar way to what you did with [reference number](https://github.com/alphagov/govuk-coronavirus-business-volunteer-form/pull/145/files#diff-ebeb2548534e3e5e7c407258f3d531bdR27), even if it's fairly lightweight, a map method has some overhead right?

app/controllers/coronavirus_form/expert_advice_type_controller.rb
```diff
@@ -38,11 +38,11 @@ def submit
...
+  OTHER = I18n.t("coronavirus_form.questions.#{PAGE}.options.other.label")
 
   def selected_other?(expert_advice_type)
-    expert_advice_type.include?(
-      I18n.t("coronavirus_form.questions.#{PAGE}.options.other.label"),
-    )
+    expert_advice_type.include?(OTHER)
```
huwd, line 45: nice

app/views/coronavirus_form/check_answers.html.erb
```diff
@@ -6,10 +6,12 @@
   <%= render "govuk_publishing_components/components/back_link", { href: previous_path } %>
 <% end %>
 
-<%= render "govuk_publishing_components/components/title", {
-  title: t('check_your_answers.title'),
-  margin_top: 0
-} %>
+<% cache(t('check_your_answers.title')) do %>
```
huwd: ah ha! [this must be the fragment caching then](https://guides.rubyonrails.org/caching_with_rails.html).

so... how do we invalidate this?
If we bumped the version of govuk_publishing_components would these get auto invalidated?
What if we added css, or a new margin_top attribute?

huwd (commented):
really interesting, will leave approval to those that know a bit more about optimization i rails, left a few questions to check I've understood what's going on here.

Once i understand it all a bit better would be happy to approve if no-one has beat me to it!

app/controllers/coronavirus_form/additional_product_controller.rb
bilbof: Actually I'm not sure what the impact of using a constant is on performance. It's just a refactor to make things clearer :)

huwd: cool, very nice!

app/controllers/coronavirus_form/check_answers_controller.rb
bilbof: Yeah, although since the method is already called just once, it won't lead to a performance improvement.

app/controllers/coronavirus_form/check_answers_controller.rb
bilbof: Sorry, it's not super clear. There's probably a better way to do this!

We have an element at a given index, and we want to puts the elements either side of that index into two arrays. For example, in an array 1-10, we want to put 1,2,3,4,5,6 into one list, and 8,9,10 into another, and exclude 7 from both:

```ruby
items = (1..10).to_a   # [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
index = 6              # 7
items.slice(0, index)  # [1, 2, 3, 4, 5, 6]
items.slice(index+1..) # [8, 9, 10]
```

app/controllers/coronavirus_form/expert_advice_type_controller.rb
bilbof: Since this is a constant `I18n.t("coronavirus_form...` will only get evaluated once when a new instance is created, I think.

app/controllers/coronavirus_form/check_answers_controller.rb
huwd: ah gotcha, no that makes sense, just not a pattern i'd seen before!
it's cool.

app/views/coronavirus_form/check_answers.html.erb
bilbof: Great question! I didn't consider the publishing components gem being updated, this would've been invalidated when the content changed, but not when the gem changed.

I've made a change to the view fragment cache keys, following what DDH recommends in [this blog post](https://signalvnoise.com/posts/3113-how-key-based-cache-expiration-works), by adding versions in.

Now we can invalidate the cache by changing the cache key version number when we make a change to the view

If the govuk publishing components gem version gets bumped, or the text the component renders changes, this should also invalidate the cache.

bilbof:
I'm closing this as we haven't had latency issues. We can reopen / redo this work should we need to improve the response times.

PR #147, opened 2020-03-30
Enable users to remove products

bilbof:
On the check your answers page users will be able to remove the products that they previously added.

Screenshot 2020-03-30 at 10 35 48

https://trello.com/c/bRb6bBS7/161-add-option-to-remove-product-to-check-your-answers-screen

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -24,6 +24,7 @@ Rails.application.routes.draw do
   post "/are-you-a-manufacturer" => "coronavirus_form/are_you_a_manufacturer#submit"
 
   get "/product-details" => "coronavirus_form/product_details#show"
+  get "/product-details/:id/delete" => "coronavirus_form/product_details#destroy"
   post "/product-details" => "coronavirus_form/product_details#submit"
 
   get "/additional-product" => "coronavirus_form/additional_product#show"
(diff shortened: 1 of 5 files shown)


config/routes.rb
```diff
@@ -24,6 +24,7 @@
   post "/are-you-a-manufacturer" => "coronavirus_form/are_you_a_manufacturer#submit"
 
   get "/product-details" => "coronavirus_form/product_details#show"
+  get "/product-details/:id/delete" => "coronavirus_form/product_details#destroy"
```
brucebolt, line 27: Can we make this a DELETE request?

brucebolt (approved):
Looks good.  I've just left a comment to see if we can change this from a GET to a DELETE request type?

config/routes.rb
bilbof: I would like that; as far as I can see the component only supports GETs.

https://components.publishing.service.gov.uk/component-guide/summary_list/with_edit_on_individual_items

brucebolt: That's a shame

bilbof: One option would be to add an intermediary page, with "are you sure you want to delete this product?" with a form that makes a DELETE request. I'm not sure of the value of this for users, though.

brucebolt: I'd just go for the MVP approach at the moment (with GET request)

alex-ju:
@bilbof thank you for getting this one done! 🎉 As you mentioned, it seems like in a 2/3 container the Change/Delete actions collapse. I'll raise a tiny PR with a CSS override in this app to cope with it.

bilbof:
@alex-ju thank you, you rock 🏅

PR #149, opened 2020-03-30
Store medical equipment type in session only when valid

bilbof:
This prevents one from storing an invalid value for this field in the session.

https://trello.com/c/zIOCiJLn/143


brucebolt (approved):
👍

PR #152, opened 2020-03-30
Fix Concourse pipeline

brucebolt:
The resource to run a task daily was added to the wrong section.  Moving to resources so the pipeline can be deployed correctly.


issyl0 (approved):
oops

bilbof (approved):

PR #154, opened 2020-03-30
Remove full stops from hint text

brucebolt:
Full stops should be used at the end of example.  Full stops should not be used for lead-ins to a question.  Therefore removing full stops from the end of lead-ins to questions.

Also took this as an opportunity to make the quoting consistent throughout the internationalisation YAML file.

Trello card: https://trello.com/c/ur7pxeNs


issyl0 (approved):
😍

PR #156, opened 2020-03-30
Allow user to indicate transport capacity

brucebolt:
Trello: https://trello.com/c/A1S8pkM8/103-indicate-transport-capacity

Adds an "other" option and also a textarea to gather details of the transport that can be offered.  Check your answers page updated to display both fields alongside the same question.

Also refactors the mandatory text field validator to keep it consistent with the other validators.

Screenshot 2020-03-30 at 15 42 35

Screenshot 2020-03-30 at 15 44 08

Trello card: https://trello.com/c/A1S8pkM8

### DIFF
--- a/app/controllers/coronavirus_form/business_details_controller.rb
+++ b/app/controllers/coronavirus_form/business_details_controller.rb
@@ -31,7 +31,7 @@ private
   PAGE = "business_details"
 
   def validate_fields(business_details)
-    missing_fields = validate_mandatory_text_fields(REQUIRED_FIELDS, PAGE)
+    missing_fields = validate_mandatory_text_fields(PAGE, REQUIRED_FIELDS)
     company_size_validation = validate_radio_field("#{PAGE}.company_size", radio: business_details["company_size"])
     company_location_validation = validate_radio_field("#{PAGE}.company_location", radio: business_details["company_location"])
     missing_fields + company_size_validation + company_location_validation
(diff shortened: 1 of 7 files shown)


app/controllers/coronavirus_form/business_details_controller.rb
```diff
@@ -31,7 +31,7 @@ def submit
   PAGE = "business_details"
 
   def validate_fields(business_details)
-    missing_fields = validate_mandatory_text_fields(REQUIRED_FIELDS, PAGE)
+    missing_fields = validate_mandatory_text_fields(PAGE, REQUIRED_FIELDS)
```
huwd, line 34: alphabetising?

app/controllers/coronavirus_form/check_answers_controller.rb
```diff
@@ -112,6 +116,20 @@ def product_details(products)
     end
   end
 
+  def transport_type(question)
+    answer = []
+    answer << session["transport_type"].flatten.to_sentence
```
huwd, line 121: [to_sentence is cool](https://apidock.com/rails/Array/to_sentence). Hadn't seen that before!

app/views/coronavirus_form/transport_type.html.erb
```diff
@@ -27,5 +28,15 @@
     }
   end
 } %>
+
+<%= render "govuk_publishing_components/components/textarea", {
```
huwd, line 32: @bilbof has an open PR to cache some of these, if that gets merged first we should adjust these here too?

huwd (approved):

app/controllers/coronavirus_form/business_details_controller.rb
brucebolt: Yes.  Also the other validator methods all take page first and the other arguments next.

PR #157, opened 2020-03-30
Remove PAGE and NEXT_PAGE constants

bilbof:
This copies the work done in https://github.com/alphagov/govuk-coronavirus-vulnerable-people-form/pull/199/commits/8105c2ebfc1f097657a3ce394dda7aef2aef3997.

### DIFF
--- a/app/controllers/coronavirus_form/additional_product_controller.rb
+++ b/app/controllers/coronavirus_form/additional_product_controller.rb
@@ -1,35 +1,27 @@
 # frozen_string_literal: true
 
 class CoronavirusForm::AdditionalProductController < ApplicationController
-  before_action :check_first_question_answered, only: :show
-
-  def show
-    render "coronavirus_form/#{PAGE}"
-  end
-
   def submit
     additional_product = strip_tags(params[:additional_product]).presence
 
     invalid_fields = validate_radio_field(
-      PAGE,
+      controller_name,
       radio: additional_product,
     )
 
     if invalid_fields.any?
       flash.now[:validation] = invalid_fields
-      render "coronavirus_form/#{PAGE}"
+      render controller_path
     elsif additional_product == I18n.t("coronavirus_form.questions.additional_product.options.option_yes.label")
-      redirect_to controller: "coronavirus_form/medical_equipment_type", action: "show"
+      redirect_to medical_equipment_type_path
     elsif session["check_answers_seen"]
-      redirect_to controller: "coronavirus_form/check_answers", action: "show"
+      redirect_to check_your_answers_path
     else
-      redirect_to controller: "coronavirus_form/hotel_rooms", action: "show"
+      redirect_to hotel_rooms_path
     end
   end
 
 private
-
-  PAGE = "additional_product"
 
   def previous_path
     session["product_details"] ||= []
--- a/app/controllers/coronavirus_form/additional_product_controller.rb
+++ b/app/controllers/coronavirus_form/additional_product_controller.rb
@@ -13,11 +13,11 @@ class CoronavirusForm::AdditionalProductController < ApplicationController
       flash.now[:validation] = invalid_fields
       render controller_path
     elsif additional_product == I18n.t("coronavirus_form.questions.additional_product.options.option_yes.label")
-      redirect_to medical_equipment_type_path
+      redirect_to medical_equipment_type_url
     elsif session["check_answers_seen"]
-      redirect_to check_your_answers_path
+      redirect_to check_your_answers_url
     else
-      redirect_to hotel_rooms_path
+      redirect_to hotel_rooms_url
     end
   end
 
@@ -26,6 +26,6 @@ private
   def previous_path
     session["product_details"] ||= []
     latest_product_id = (session["product_details"].last || {}).dig("product_id")
-    "/product-details?product_id=#{latest_product_id}"
+    product_details_url(product_id: latest_product_id)
   end
 end
--- a/app/controllers/coronavirus_form/are_you_a_manufacturer_controller.rb
+++ b/app/controllers/coronavirus_form/are_you_a_manufacturer_controller.rb
@@ -15,15 +15,15 @@ class CoronavirusForm::AreYouAManufacturerController < ApplicationController
       flash.now[:validation] = invalid_fields
       render controller_path
     elsif session["check_answers_seen"]
-      redirect_to check_your_answers_path
+      redirect_to check_your_answers_url
     else
-      redirect_to medical_equipment_type_path
+      redirect_to medical_equipment_type_url
     end
   end
 
 private
 
   def previous_path
-    medical_equipment_path
+    medical_equipment_url
   end
 end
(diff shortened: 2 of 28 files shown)


app/controllers/coronavirus_form/additional_product_controller.rb
```diff
@@ -1,36 +1,30 @@
 # frozen_string_literal: true
 
 class CoronavirusForm::AdditionalProductController < ApplicationController
-  before_action :check_first_question_answered, only: :show
-
-  def show
-    render "coronavirus_form/#{PAGE}"
-  end
+  before_action :check_first_question_answered, except: :submit
```
leenagupte: Why has this change from an "only" to an "except"?

If you left this as an "only" it could be moved into the application controller. e.g. https://github.com/alphagov/govuk-coronavirus-vulnerable-people-form/blob/master/app/controllers/application_controller.rb#L9

app/controllers/coronavirus_form/additional_product_controller.rb
```diff
@@ -1,36 +1,30 @@
...
       radio: additional_product,
     )
 
     if invalid_fields.any?
       flash.now[:validation] = invalid_fields
-      render "coronavirus_form/#{PAGE}"
+      render controller_path
     elsif additional_product == I18n.t("coronavirus_form.questions.additional_product.options.option_yes.label")
-      redirect_to controller: "coronavirus_form/medical_equipment_type", action: "show"
+      redirect_to medical_equipment_type_path
```
leenagupte: Nitpick: The vulnerable people form is using the `_url` helper rather than `_path`. There's an explanation for why in the [PR description](https://github.com/alphagov/govuk-coronavirus-vulnerable-people-form/pull/199)

app/controllers/coronavirus_form/are_you_a_manufacturer_controller.rb
```diff
@@ -1,37 +1,35 @@
 # frozen_string_literal: true
 
 class CoronavirusForm::AreYouAManufacturerController < ApplicationController
-  before_action :check_first_question_answered, only: :show
+  before_action :check_first_question_answered, except: :submit
 
   def show
     session["are_you_a_manufacturer"] ||= []
-    render "coronavirus_form/#{PAGE}"
+    super
```
leenagupte: I think if you use `session.dig("are_you_a_manufacturer") you might not need to memoize it here, and then could remove the method.

app/controllers/coronavirus_form/business_details_controller.rb
```diff
@@ -1,39 +1,32 @@
...
+    invalid_fields = validate_field_response_length(controller_name, TEXT_FIELDS) +
       validate_fields(@business_details)
 
     if invalid_fields.any?
       flash.now[:validation] = invalid_fields
-      render "coronavirus_form/#{PAGE}"
+      render controller_path
     else
-      redirect_to controller: session["check_answers_seen"] ? "coronavirus_form/check_answers" : "coronavirus_form/#{NEXT_PAGE}", action: "show"
+      redirect_to session["check_answers_seen"] ? check_your_answers_path : contact_details_path
```
leenagupte: Oh dear! I know this isn't part of this PR, but inconsistent use of a ternary operation. The other controllers use an `elsif`.

app/controllers/coronavirus_form/check_answers_controller.rb
```diff
@@ -103,7 +102,7 @@ def product_details(products)
         field: "Details for product #{product['product_name']}",
         value: sanitize(prod.join(joiner)),
         edit: {
-          href: "/product-details?product_id=#{product['product_id']}",
+          href: product_details_path(product_id: product["product_id"]),
```
leenagupte: Nice!

app/controllers/coronavirus_form/medical_equipment_type_controller.rb
```diff
@@ -1,34 +1,34 @@
...
 
     if invalid_fields.any?
       flash.now[:validation] = invalid_fields
-      render "coronavirus_form/#{PAGE}"
+      render controller_path
     else
       add_product_to_session(@product)
       redirect_to(
-        controller: "coronavirus_form/#{NEXT_PAGE}",
+        controller: "coronavirus_form/product_details",
```
leenagupte: This could be changed to:

```
redirect_to product_details_url(product_id: @product["product_id"])
```

leenagupte (commented):
This is looking great!

I've added some inline comments/suggestions. I've only added them in once place, they they apply across the board.

app/controllers/coronavirus_form/additional_product_controller.rb
bilbof: I avoided putting this into the application controller because I think pages like `/privacy` shouldn't have this check.

I've changed this to be an except because rubocop doesn't like specifying a before_action on an inherited method.

leenagupte: So privacy can be excluded with `skip_before_action`,
e.g. https://github.com/alphagov/govuk-coronavirus-vulnerable-people-form/blob/master/app/controllers/coronavirus_form/privacy_controller.rb

You could then `before_action` to the application controller, so it doesn't have to be repeated in all the forms.

app/controllers/coronavirus_form/additional_product_controller.rb
bilbof: Interesting, thanks 👍

bilbof:
Thanks @leenagupte I've pushed some changes.

app/controllers/coronavirus_form/additional_product_controller.rb
```diff
@@ -1,36 +1,28 @@
...
-      redirect_to controller: "coronavirus_form/hotel_rooms", action: "show"
+      redirect_to hotel_rooms_path
     end
   end
 
 private
 
-  PAGE = "additional_product"
-
   def previous_path
```
leenagupte, line 26: Can the `path` helper be used in this method too? 
e.g. 
`product_details_path(product_id: latest_product_id)`
rather than:
`"/product-details?product_id=#{latest_product_id}"`

app/controllers/coronavirus_form/product_details_controller.rb
```diff
@@ -99,8 +99,8 @@ def sanitized_product(params)
   end
 
   def previous_path
-    return check_your_answers_path if session["check_answers_seen"]
+    return check_your_answers_url if session["check_answers_seen"]
```
leenagupte, line 102: This check threw me a bit, because it's not on the other pages, but I guess it's because this is a looping question?

leenagupte (approved):
Looks good! 👍 

I had a minor suggestion and a question, but nothing blocking.

app/controllers/coronavirus_form/product_details_controller.rb
bilbof: Yeah I think it's intentional.

app/controllers/coronavirus_form/additional_product_controller.rb
bilbof: Thanks, I missed that.

PR #158, opened 2020-03-30
Return if transport type question has not been answered

brucebolt:
Otherwise we are attempting to flatten nil if the session value for transport_type has not been set.

### DIFF
--- a/app/controllers/coronavirus_form/check_answers_controller.rb
+++ b/app/controllers/coronavirus_form/check_answers_controller.rb
@@ -47,7 +47,7 @@ private
         end
 
         if question.eql?("transport_type")
-          next transport_type(question)
+          next transport_type
         end
 
         value = case session[question]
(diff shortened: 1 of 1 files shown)


app/controllers/coronavirus_form/check_answers_controller.rb
```diff
@@ -117,6 +117,8 @@ def product_details(products)
   end
 
   def transport_type(question)
+    return [] unless session[question]
```
bilbof: Looks good. Since question will always be "transport_type", perhaps we could remove the arguments for this method and write `session["transport_type"]` here.

One could also do `Array(session["transport_type"]).flatten.to_sentence` below

bilbof (approved):

app/controllers/coronavirus_form/check_answers_controller.rb
brucebolt: Good points.  Have updated.

bilbof (approved):

PR #166, opened 2020-04-01
Update care qualification question

brucebolt:
Adds an additional question asking what type of care can be offered.  Additionally adds an 'Or' separator into the list of qualifications.

Screenshot 2020-04-01 at 11 59 07

Trello card: https://trello.com/c/i1bH0AtM

### DIFF
--- a/app/views/coronavirus_form/offer_care_qualifications.erb
+++ b/app/views/coronavirus_form/offer_care_qualifications.erb
@@ -54,7 +54,13 @@
           error_message: error_items("offer_care_qualifications_type"),
           value: session[:offer_care_qualifications_type],
         })
-      },
+      }
+    ]
+  } %>
+  <%= tag.p t("coronavirus_form.questions.offer_care_qualifications.care_qualifications.or"), class: "govuk-body" %>
+  <%= render "govuk_publishing_components/components/checkboxes", {
+    name: "offer_care_qualifications[]",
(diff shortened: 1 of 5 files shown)


issyl0 (approved):

app/views/coronavirus_form/offer_care_qualifications.erb
```diff
@@ -54,7 +54,13 @@
           error_message: error_items("offer_care_qualifications_type"),
           value: session[:offer_care_qualifications_type],
         })
-      },
+      }
+    ]
+  } %>
+  <%= tag.p t("coronavirus_form.questions.offer_care_qualifications.care_qualifications.or"), class: "govuk-body" %>
```
leenagupte, line 60: The radio button component has an [or divider](https://components.publishing.service.gov.uk/component-guide/radio/with_or_divider). Perhaps something similar could be added the checkbox component in the future? What do you think @alex-ju?

alex-ju: I added the _or_ divider to the radios component as it's an established pattern (and it makes sense to have it as a separator since the logic for radio is an or-based), but for checkboxes I'm reluctant to introduce something similar as they're and-based and introducing an _or_ would make things confusing. This is the reason we have this workaround in the app – to avoid making it available in other places.

PR #173, opened 2020-04-01
Fix error state on care qualifications

alex-ju:
Adjust ids to match the generated error summary links. Fix component attribute (for checkboxes it's `error`, not `error_message` – annoyingly inconsistent with radios).

Follow up on #166.


brucebolt (approved):
Oops!

alex-ju:
Will raise a PR in `govuk_publishing_components` to allow both `error` and `error_message` on checkboxes for backwards compatibility and our sanity.

PR #175, opened 2020-04-01
Remove communications from list of expertise

brucebolt:
This option is no longer required in the form.

Trello card: https://trello.com/c/7jtXEZ2B


alex-ju (approved):
Easy come, easy go.

PR #178, opened 2020-04-02
Add timeout page to the service and increase session expiry time

brucebolt:
We have a short expiry time on session data so user responses are not stored in the user does not submit the form.

This makes the following changes:
- Shows a pretty error when the session expires
- Increases the session timeout from 2 to 4 hours (increasing the time allowed to complete the form)

Example of session expired page:
Screenshot 2020-04-02 at 11 28 56

This mirrors changes made in https://github.com/alphagov/govuk-coronavirus-vulnerable-people-form/pull/189 and https://github.com/alphagov/govuk-coronavirus-vulnerable-people-form/pull/186 for the vulnerable people form.

Trello card: https://trello.com/c/WY2eN2JL

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -7,6 +7,8 @@ class ApplicationController < ActionController::Base
   include FormFlowHelper
   include ProductHelper
 
+  rescue_from ActionController::InvalidAuthenticityToken, with: :session_expired
+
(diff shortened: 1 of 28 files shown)


leenagupte (approved):
Looks familiar! 🤔

brucebolt:
Lifted and shifted to this app 😁

PR #180, opened 2020-04-02
Annotate route file

huwd:
A part of: [If a user goes back and chooses "no" to a question they previously chose "yes" to, remove the other data held against that question](https://trello.com/c/AnGnIZ7I/15-if-a-user-goes-back-and-chooses-no-to-a-question-they-previously-chose-yes-to-remove-the-other-data-held-against-that-question)

## What
Annotates routes with question numbers

## Why
It makes it to grok which questions are dependent on others
To work out the back button behavior (not storing now-irrelevant answers) we need to identify dependencies. This helped me do that.


brucebolt (approved):
👍

PR #181, opened 2020-04-02
Remove full stops from form field IDs

brucebolt:
We have some form fields which contain a full stop.  This breaks conditional reveals since the space is not escaped in the CSS selector.

Therefore replacing all full stops in the IDs with underscores and updating the validators to replace full stops in field names.  This maintains the link to specific errors that appear in the flash box.

Trello card: https://trello.com/c/JOdhNGhY


issyl0 (approved):
Good catch! How?!

brucebolt:
Just kept rolling back commits on the view (that had the broken conditional reveal) until it started working again.

PR #187, opened 2020-04-03
Add asset sync to copy assets to S3

richardTowers:
Same idea as https://github.com/alphagov/govuk-coronavirus-vulnerable-people-form/pull/208

https://trello.com/c/cfGnjdZV/27-fix-issues-with-assets-404ing-during-rolling-deployments

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -4,8 +4,10 @@ ruby File.read(".ruby-version").strip
 
 source "https://rubygems.org"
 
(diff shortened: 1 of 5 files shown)


richardTowers:
We've already created the S3 buckets in staging and prod for this, and we think the secret access keys are already uploaded to concourse (unless someone removed them when we were cleaning up).

bilbof (approved):

PR #191, opened 2020-04-03
Add GA to all business form pages

bilbof:
This adds Google Analytics to the business form, to help us understand how businesses are using the service to offer help, so the development team can make improvements.

We have no plans to add this to the vulnerable people service.

The GA views have been set up by John B already.

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -44,5 +44,13 @@
       }
     } %>
     <%= javascript_include_tag "application" %>
+    <!-- Global site tag (gtag.js) - Google Analytics -->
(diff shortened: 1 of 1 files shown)


richardTowers:
Do we need some kind of consent mechanism?

bilbof:
Currently there's no way to opt-out without using something like [privacy badger](https://privacybadger.org/). I believe this usually would require opt-in, but there may have been changes to restrictions for services related to COVID-19.

I've asked about this and updating the privacy policy - I'll update you when I get a response.

bilbof:
Thanks for catching this, @richardTowers. We do need to make this opt-in, so I will re-open this PR once there's a cookies page and opt-in mechanism.

PR #192, opened 2020-04-03
Return product quantity input field to default width

brucebolt:
This field was originally reduced in width with the intention that users would enter solely a number.  However, in practice users have entered additional essential information (e.g. "per week").  Making the field wider will make it easier for users to enter this information.

Before:
Screenshot 2020-04-03 at 21 36 26

After:
Screenshot 2020-04-03 at 21 35 56

Trello card: https://trello.com/c/prOOYLSB


issyl0 (approved):
It is of type "number" though and the hint text only suggests numbers. But I suppose the qualifier is good and useful for those deciding.

PR #194, opened 2020-04-06
Only show the questions people have answered on check answers page

leenagupte:
Trello: https://trello.com/c/XTtEutBv

# Expected changes
## Before
before-screenshot-govuk-coronavirus-business-volunteer-form-stg cloudapps digital-2020 04 06-10_22_46

## After
after-screenshot-localhost_5000-2020 04 06-10_31_02

### DIFF
--- a/app/helpers/check_answers_helper.rb
+++ b/app/helpers/check_answers_helper.rb
@@ -17,7 +17,7 @@ module CheckAnswersHelper
       next if question.eql?("medical_equipment_type")
       next if skip_question?(question)
 
-      next product_details(session[question]) if question.eql?("product_details")
+      next product_details if question.eql?("product_details")
       next transport_type if question.eql?("transport_type")
       next offer_care_qualifications if question.eql?("offer_care_qualifications")
 
@@ -56,7 +56,8 @@ module CheckAnswersHelper
     items.select.with_index { |_, index| index > additional_product_index }
   end
 
-  def product_details(products)
+  def product_details
+    products = session["product_details"]
     return unless products && products.any?
 
     products.map do |product|
--- a/app/helpers/check_answers_helper.rb
+++ b/app/helpers/check_answers_helper.rb
@@ -17,17 +17,9 @@ module CheckAnswersHelper
       next if question.eql?("medical_equipment_type")
       next if skip_question?(question)
 
-      if question.eql?("product_details")
-        next product_details(session[question])
-      end
-
-      if question.eql?("transport_type")
-        next transport_type
-      end
-
-      if question.eql?("offer_care_qualifications")
-        next offer_care_qualifications
-      end
+      next product_details(session[question]) if question.eql?("product_details")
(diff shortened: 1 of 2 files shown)


app/helpers/check_answers_helper.rb
```diff
@@ -1,10 +1,21 @@
...
+    offer_space_type
+    offer_care_qualifications
+  ).freeze
+
   def items
     questions.map { |question|
       # We have answers as strings and hashes. The hashes need a little more
       # work to make them readable.
 
       next if question.eql?("medical_equipment_type")
```
JonathanHallam: Do we need to keep this given that it's now in skippable questions? Am I reading it right that under the old system this just wouldn't have shown up in the answers?

app/helpers/check_answers_helper.rb
```diff
@@ -37,6 +48,10 @@ def items
     }.compact.flatten
   end
 
+  def skip_question?(question)
+    question.in?(SKIPPABLE_QUESTIONS) && session[question].blank?
```
JonathanHallam, line 44: May not be the right forum for this but I'm going to ask questions given my newity! Does a user have to explicitly skip or just not enter any info and click continue?

app/helpers/check_answers_helper.rb
```diff
@@ -17,17 +17,9 @@ def items
...
-      if question.eql?("transport_type")
-        next transport_type
-      end
-
-      if question.eql?("offer_care_qualifications")
-        next offer_care_qualifications
-      end
+      next product_details(session[question]) if question.eql?("product_details")
+      next transport_type if question.eql?("transport_type")
+      next offer_care_qualifications if question.eql?("offer_care_qualifications")
```
JonathanHallam, line 20: nice 👍

JonathanHallam (approved):
Left a few questions but code changes look good :)

app/helpers/check_answers_helper.rb
leenagupte: You're right! I'll go and update this.

app/helpers/check_answers_helper.rb
leenagupte: They don't see these questions at all. If they answer "no" to one of the preceding questions, they jump right over them. 

e.g. https://github.com/alphagov/govuk-coronavirus-business-volunteer-form/blob/master/app/controllers/coronavirus_form/hotel_rooms_controller.rb#L17-L23

If they say "no" to any hotel rooms, they jump straight to the transport question and are not asked how many hotel rooms they have.

app/helpers/check_answers_helper.rb
leenagupte: @JonathanHallam updated here: https://github.com/alphagov/govuk-coronavirus-business-volunteer-form/pull/194/commits/fcf6cd5123c842650c59664bd9ca7d1795cef095

PR #197, opened 2020-04-06
Perform validations before setting session values

brucebolt:
We are currently saving invalid responses to the session hash before doing validations.  We need to validate first, but still replay the invalid response back to the user.  This also removes all use of the session hash out of the views.

As a side effect of this change we have resolved all string vs symbols issues for the stored values, using symbols in all places now.

This change was made for the vulnerable people registration service in https://github.com/alphagov/govuk-coronavirus-vulnerable-people-form/pull/230.

Whilst this looks like a lot of lines changed, it is mostly the same amendment but in multiple places.

Trello card: https://trello.com/c/PygFKz9a

### DIFF
--- a/app/controllers/coronavirus_form/contact_details_controller.rb
+++ b/app/controllers/coronavirus_form/contact_details_controller.rb
@@ -4,44 +4,44 @@ class CoronavirusForm::ContactDetailsController < ApplicationController
   REQUIRED_FIELDS = %w(contact_name phone_number).freeze
   TEXT_FIELDS = %w(contact_name role phone_number email).freeze
 
-  def show
-    session[:contact_details] ||= {}
-    super
-  end
-
   def submit
-    session[:contact_details] ||= {}
-    session[:contact_details]["contact_name"] = strip_tags(params[:contact_name]).presence
-    session[:contact_details]["role"] = strip_tags(params[:role]).presence
-    session[:contact_details]["phone_number"] = strip_tags(params[:phone_number]).presence
-    session[:contact_details]["email"] = strip_tags(params[:email]).presence
+    @form_responses = {
+      contact_details: {
+        contact_name: strip_tags(params[:contact_name]).presence,
+        role: strip_tags(params[:role]).presence,
+        phone_number: strip_tags(params[:phone_number]).presence,
+        email: strip_tags(params[:email]).presence,
+      },
+    }
 
-    invalid_fields = validate_field_response_length(controller_name, TEXT_FIELDS) +
-      validate_fields(session[:contact_details])
+    invalid_fields = validate_fields
 
     if invalid_fields.any?
       flash.now[:validation] = invalid_fields
       log_validation_error(invalid_fields)
       render controller_path
     else
+      session[:contact_details] = @form_responses[:contact_details]
       redirect_to check_your_answers_url
     end
   end
 
 private
 
-  def validate_fields(contact_details)
-    missing_fields = validate_missing_fields(contact_details)
-    email_validation = validate_email_address("email", contact_details["email"])
-    missing_fields + email_validation
+  def validate_fields
+    [
+      validate_field_response_length(controller_name, TEXT_FIELDS),
+      validate_missing_fields,
+      validate_email_address("email", @form_responses.dig(:contact_details, :email)),
+    ].flatten.compact
   end
 
-  def validate_missing_fields(product)
+  def validate_missing_fields
     REQUIRED_FIELDS.each_with_object([]) do |field, invalid_fields|
-      next if product[field].present?
+      next if @form_responses[:contact_details][field.to_sym].present?
 
       invalid_fields << {
-        field: field.to_s,
+        field: field,
         text: t("coronavirus_form.questions.#{controller_name}.#{field}.custom_error",
                 default: t("coronavirus_form.errors.missing_mandatory_text_field",
                            field: t("coronavirus_form.questions.#{controller_name}.#{field}.label")).humanize),
--- a/app/controllers/coronavirus_form/expert_advice_type_controller.rb
+++ b/app/controllers/coronavirus_form/expert_advice_type_controller.rb
@@ -4,21 +4,17 @@ class CoronavirusForm::ExpertAdviceTypeController < ApplicationController
   TEXT_FIELDS = %w(expert_advice_type_other).freeze
 
   def submit
-    expert_advice_type = Array(params[:expert_advice_type]).map { |item| strip_tags(item).presence }.compact
-    expert_advice_type_other = strip_tags(params[:expert_advice_type_other]).presence
-    session[:expert_advice_type] = expert_advice_type
-    session[:expert_advice_type_other] = if selected_other?(expert_advice_type)
-                                           expert_advice_type_other
-                                         else
-                                           ""
-                                         end
+    @form_responses = {
+      expert_advice_type: Array(params[:expert_advice_type]).map { |item| strip_tags(item).presence }.compact,
+    }
+    @form_responses[:expert_advice_type_other] = strip_tags(params[:expert_advice_type_other]).presence if selected_other?
 
     invalid_fields = validate_field_response_length(controller_name, TEXT_FIELDS) +
       validate_checkbox_field(
         controller_name,
-        values: expert_advice_type,
+        values: @form_responses[:expert_advice_type],
         allowed_values: I18n.t("coronavirus_form.questions.#{controller_name}.options").map { |_, item| item.dig(:label) },
-        other: expert_advice_type_other,
+        other: @form_responses[:expert_advice_type_other],
       )
 
     if invalid_fields.any?
@@ -26,16 +22,23 @@ class CoronavirusForm::ExpertAdviceTypeController < ApplicationController
       log_validation_error(invalid_fields)
       render controller_path
     elsif session["check_answers_seen"]
+      update_session_store
       redirect_to check_your_answers_url
     else
+      update_session_store
       redirect_to offer_care_url
     end
   end
 
 private
 
-  def selected_other?(expert_advice_type)
-    expert_advice_type.include?(
+  def update_session_store
+    session[:expert_advice_type] = @form_responses[:expert_advice_type]
+    session[:expert_advice_type_other] = @form_responses[:expert_advice_type_other]
+  end
+
+  def selected_other?
+    @form_responses[:expert_advice_type].include?(
       I18n.t("coronavirus_form.questions.#{controller_name}.options.other.label"),
     )
   end
--- a/app/controllers/coronavirus_form/medical_equipment_controller.rb
+++ b/app/controllers/coronavirus_form/medical_equipment_controller.rb
@@ -26,8 +26,6 @@ class CoronavirusForm::MedicalEquipmentController < ApplicationController
     elsif @form_responses[:medical_equipment] == I18n.t("coronavirus_form.questions.#{controller_name}.options.option_no.label")
       update_session_store
       redirect_to hotel_rooms_url
-    else
-      render controller_path
     end
   end
 
--- a/app/controllers/coronavirus_form/medical_equipment_controller.rb
+++ b/app/controllers/coronavirus_form/medical_equipment_controller.rb
@@ -4,13 +4,13 @@ class CoronavirusForm::MedicalEquipmentController < ApplicationController
   skip_before_action :check_first_question_answered
 
   def submit
-    medical_equipment = strip_tags(params[:medical_equipment]).presence
-
-    session[:medical_equipment] = medical_equipment
+    @form_responses = {
+      medical_equipment: strip_tags(params[:medical_equipment]).presence,
+    }
 
     invalid_fields = validate_radio_field(
       controller_name,
-      radio: medical_equipment,
+      radio: @form_responses[:medical_equipment],
     )
 
     if invalid_fields.any?
@@ -18,12 +18,13 @@ class CoronavirusForm::MedicalEquipmentController < ApplicationController
       log_validation_error(invalid_fields)
       render controller_path
     elsif session["check_answers_seen"]
+      update_session_store
       redirect_to check_your_answers_url
-    elsif session[:medical_equipment] == I18n.t("coronavirus_form.questions.#{controller_name}.options.option_yes.label")
-      # YES goes to what kind of business are you / are_you_a_manufacturer
+    elsif @form_responses[:medical_equipment] == I18n.t("coronavirus_form.questions.#{controller_name}.options.option_yes.label")
+      update_session_store
       redirect_to are_you_a_manufacturer_url
-    elsif session[:medical_equipment] == I18n.t("coronavirus_form.questions.#{controller_name}.options.option_no.label")
-      # NO goes to hotel rooms
+    elsif @form_responses[:medical_equipment] == I18n.t("coronavirus_form.questions.#{controller_name}.options.option_no.label")
+      update_session_store
       redirect_to hotel_rooms_url
     else
       render controller_path
@@ -31,6 +32,10 @@ class CoronavirusForm::MedicalEquipmentController < ApplicationController
   end
 
 private
+
+  def update_session_store
+    session[:medical_equipment] = @form_responses[:medical_equipment]
+  end
 
   def previous_path
     "/"
--- a/app/controllers/coronavirus_form/offer_care_qualifications_controller.rb
+++ b/app/controllers/coronavirus_form/offer_care_qualifications_controller.rb
@@ -4,25 +4,23 @@ class CoronavirusForm::OfferCareQualificationsController < ApplicationController
   TEXT_FIELDS = %w(offer_care_qualifications_type).freeze
 
   def submit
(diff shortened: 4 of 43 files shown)


app/controllers/coronavirus_form/contact_details_controller.rb
```diff
@@ -4,41 +4,41 @@ class CoronavirusForm::ContactDetailsController < ApplicationController
...
+      validate_field_response_length(controller_name, TEXT_FIELDS),
+      validate_missing_fields(@form_responses[:contact_details]),
+      validate_email_address("email", @form_responses.dig(:contact_details, :email)),
+    ].flatten.compact
   end
 
   def validate_missing_fields(product)
     REQUIRED_FIELDS.each_with_object([]) do |field, invalid_fields|
-      next if product[field].present?
+      next if product[field.to_sym].present?
```
leenagupte: It's odd that you're doing a `.to_sym` here and a `.to_s` on line 44.  Perhaps `REQUIRED_FIELDS` could be made into a list of labels so that only one is needed?

app/controllers/coronavirus_form/contact_details_controller.rb
```diff
@@ -4,41 +4,41 @@ class CoronavirusForm::ContactDetailsController < ApplicationController
   REQUIRED_FIELDS = %w(contact_name phone_number).freeze
   TEXT_FIELDS = %w(contact_name role phone_number email).freeze
 
-  def show
-    session[:contact_details] ||= {}
-    super
-  end
```
leenagupte, line 10: 👏

app/controllers/coronavirus_form/expert_advice_type_controller.rb
```diff
@@ -4,30 +4,30 @@ class CoronavirusForm::ExpertAdviceTypeController < ApplicationController
...
-    session[:expert_advice_type] = expert_advice_type
-    session[:expert_advice_type_other] = if selected_other?(expert_advice_type)
-                                           expert_advice_type_other
-                                         else
-                                           ""
-                                         end
+    @form_responses = {
+      expert_advice_type: Array(params[:expert_advice_type]).map { |item| strip_tags(item).presence }.compact,
+    }
+    @form_responses[:expert_advice_type_other] = strip_tags(params[:expert_advice_type_other]).presence if selected_other?(@form_responses[:expert_advice_type])
```
leenagupte: Now that an instance variable is being used, does `@form_responses[:expert_advice_type]` still need to be passed to `selected_other?`

app/controllers/coronavirus_form/medical_equipment_controller.rb
```diff
@@ -4,26 +4,27 @@ class CoronavirusForm::MedicalEquipmentController < ApplicationController
...
+    elsif @form_responses[:medical_equipment] == I18n.t("coronavirus_form.questions.#{controller_name}.options.option_yes.label")
+      session[:medical_equipment] = @form_responses[:medical_equipment]
       redirect_to are_you_a_manufacturer_url
-    elsif session[:medical_equipment] == I18n.t("coronavirus_form.questions.#{controller_name}.options.option_no.label")
-      # NO goes to hotel rooms
+    elsif @form_responses[:medical_equipment] == I18n.t("coronavirus_form.questions.#{controller_name}.options.option_no.label")
+      session[:medical_equipment] = @form_responses[:medical_equipment]
       redirect_to hotel_rooms_url
     else
       render controller_path
```
leenagupte: This is out of scope for this PR, but why is this `else` option here?

app/controllers/coronavirus_form/offer_care_qualifications_controller.rb
```diff
@@ -31,8 +29,14 @@ def submit
       log_validation_error(invalid_fields)
       render controller_path
     elsif session["check_answers_seen"]
+      session[:offer_care_type] = @form_responses[:offer_care_type]
+      session[:offer_care_qualifications] = @form_responses[:offer_care_qualifications]
+      session[:offer_care_qualifications_type] = @form_responses[:offer_care_qualifications_type]
```
leenagupte: Perhaps this could be extracted into a method?

leenagupte: Perhaps every controller could be forced to implement a `update_session_store ` method similar to OfferSpaceTypeController? That way the logic only needs to be added once, and it there would less chance of making copy and paste errors?

app/controllers/coronavirus_form/offer_space_type_controller.rb
```diff
@@ -8,34 +8,46 @@ class CoronavirusForm::OfferSpaceTypeController < ApplicationController
...
 
-    invalid_fields = validate_fields(offer_space_type)
+    invalid_fields = validate_fields
 
     if invalid_fields.any?
       flash.now[:validation] = invalid_fields
       log_validation_error(invalid_fields)
       render controller_path
     elsif session["check_answers_seen"]
+      update_session_store
```
leenagupte, line 26: Nice!

app/controllers/coronavirus_form/product_details_controller.rb
```diff
@@ -5,24 +5,25 @@ class CoronavirusForm::ProductDetailsController < ApplicationController
...
-    @product = find_product(params["product_id"], session["product_details"])
+    session[:product_details] ||= []
+    @product = find_product(params["product_id"], session[:product_details])
     super
   end
 
   def submit
-    @product = current_product(params[:product_id], session[:product_details]).merge(sanitized_product(params))
-    add_product_to_session(@product)
+    @product = current_product(params["product_id"], session[:product_details]).merge(sanitized_product(params))
```
leenagupte, line 14: Does the session need to be updated here? Can a `@form_response` be set here instead?

app/controllers/coronavirus_form/product_details_controller.rb
```diff
@@ -37,43 +38,45 @@ def destroy
...
-  def validate_fields(product)
-    missing_fields = validate_missing_fields(product)
-    product_equipment_type = selected_ppe? ? validate_radio_field("#{controller_name}.equipment_type", radio: product["equipment_type"]) : []
-    product_location_validation = validate_radio_field("#{controller_name}.product_location", radio: product["product_location"])
-    postcode_validation = validate_product_postcode(product)
-    missing_fields + product_equipment_type + product_location_validation + postcode_validation
+  def validate_fields
+    [
+      validate_field_response_length(controller_name, TEXT_FIELDS),
+      validate_missing_fields(@product),
```
leenagupte: Shouldn't need to pass in an instance variable.

app/controllers/coronavirus_form/product_details_controller.rb
```diff
@@ -37,43 +38,45 @@ def destroy
...
+      selected_ppe? ? validate_radio_field("#{controller_name}.equipment_type", radio: @product[:equipment_type]) : [],
```
leenagupte, line 56: This could just be an inline if:
```
validate_radio_field("#{controller_name}.equipment_type", radio: @product[:equipment_type]) if selected_ppe?
```
the `nil` returned by the conditional being false will be removed by the `compact` anyway.

app/controllers/coronavirus_form/product_details_controller.rb
```diff
@@ -37,43 +38,45 @@ def destroy
...
+      validate_radio_field("#{controller_name}.product_location", radio: @product[:product_location]),
+      validate_product_postcode(@product),
```
leenagupte: Also, here, the instance variable shouldn't need to be passed to a local method.

app/controllers/coronavirus_form/transport_type_controller.rb
```diff
@@ -23,8 +23,12 @@ def submit
       log_validation_error(invalid_fields)
       render controller_path
     elsif session["check_answers_seen"]
+      session[:transport_type] = @form_responses[:transport_type]
+      session[:transport_description] = @form_responses[:transport_description]
```
leenagupte: This too could benefit from being extracted into a method.

app/views/coronavirus_form/are_you_a_manufacturer.html.erb
```diff
@@ -23,7 +23,7 @@
     {
       value:  item[:label],
       label:  item[:label],
-      checked: session.fetch("are_you_a_manufacturer", []).include?(item[:label]),
+      checked: @form_responses.fetch(:$1, []).include?(item[:label]),
```
leenagupte: Oooh I haven't used `:$1` like this before. Was this change it response to the linter?

leenagupte: Can we be sure what `$1` is?

app/views/coronavirus_form/business_details.html.erb
```diff
@@ -26,7 +26,7 @@
     text: t('coronavirus_form.questions.business_details.company_name.label'),
   },
   error_message: error_items('company_name'),
-  value: session.dig("business_details", "company_name"),
+  value: @form_responses.fetch(:$1, :$2),
```
leenagupte: To be honest I think it's easier to read if the explicit values are kept in. `dig` can be used with symbols, so `@form_responses.dig(:business_details, :company_name)` would work. This applies to the other views too.

app/views/coronavirus_form/check_answers.html.erb
```diff
@@ -7,6 +7,8 @@
   <%= render "govuk_publishing_components/components/back_link", { href: previous_path } %>
 <% end %>
 
+<%= @form_responses.inspect %>
```
leenagupte: Can you add a note to the commit message about why this was needed and added here?

app/views/coronavirus_form/contact_details.html.erb
```diff
@@ -25,15 +25,15 @@
   id: "contact_name",
   name: "contact_name",
   error_message: error_items("contact_name"),
-  value: session[:contact_details]["contact_name"],
+  value: @form_responses.dig(:contact_details, :contact_name),
```
leenagupte, line 28: Nice!

spec/controllers/coronavirus_form/product_details_controller_spec.rb
```diff
@@ -100,9 +100,9 @@
   describe "POST submit" do
     before :each do
       session[session_key] = [{
-        "product_id" => product_id,
-        "medical_equipment_type" => I18n.t(
-          "coronavirus_form.medical_equipment_type.options.number_ppe.label",
+        product_id: product_id,
+        medical_equipment_type: I18n.t(
+          "coronavirus_form.questions.medical_equipment_type.options.number_testing_equipment.label",
```
leenagupte, line 105: I'm surprised the tests weren't bombing because this lookup failed. 🤔

leenagupte (commented):
Wow! This was a mammoth task! Thanks for taking it on. It's also great that we're using symbol keys everywhere now rather than the strange hybrid we were using before.

I've added some inline comments/suggestions. 
The main things I wanted to point out were where things are inconsistent. This is especially prevalent in the views. In some places we're using `dig`, other `fetch` and whilst I don't have preference on that, I would rather we explicitly use the key names across the board, than than `$1` and `$2` as we are in a few places. 

It would also be nice if we could consistently use an `update_session_store` method in the controllers.

Other than that it's looking really good!

app/views/coronavirus_form/business_details.html.erb
brucebolt: 🤦‍♂ This is an epic find and replace fail

app/controllers/coronavirus_form/contact_details_controller.rb
brucebolt: Good point, I hadn't seen the `to_s` several lines below.  I've removed that.  `field` is used more times as a string, so I've kept the one occurance of `to_sym`.

app/controllers/coronavirus_form/offer_care_qualifications_controller.rb
brucebolt: Good idea, I've put a `update_session_store` method into each controller that updates the session store.

app/controllers/coronavirus_form/product_details_controller.rb
brucebolt: The session isn't being updated here.  Because of the nature of products (and the user being able to add an unspecified number of them), we have a `ProductHelper` which has a `add_product_to_session` method that deals with sessions being updated.

app/controllers/coronavirus_form/product_details_controller.rb
brucebolt: Good spot.  I'd removed this from other controllers, but must've missed this one.

app/views/coronavirus_form/are_you_a_manufacturer.html.erb
brucebolt: This was a find and replace fail (which only failed in some of the views).  Have fixed.

app/views/coronavirus_form/check_answers.html.erb
brucebolt: This line shouldn't have been committed.  Have deleted.

spec/controllers/coronavirus_form/product_details_controller_spec.rb
brucebolt: We were updating the session store before doing any validation, so the test for updating the session variables always passed.

app/controllers/coronavirus_form/expert_advice_type_controller.rb
brucebolt: It doesn't.  Have removed.

app/controllers/coronavirus_form/medical_equipment_controller.rb
brucebolt: I can't see any use for it.  Have added an additional commit to remove.

app/controllers/coronavirus_form/product_details_controller.rb
brucebolt: Unfortunately the linter doesn't permit inline if statements in an array.

brucebolt:
Thank you for the review @leenagupte.  I have made the proposed changes, including an `update_session_store` method in all relevant controllers.

The `$1` and `$2` came from a faulty regular expression find-and-replace 🤦‍♂.  Didn't notice these when I reviewed the code (probably because all the lines blended into one with so many changes).

I'm hesitant to change the `dig` and `fetch` at the moment, since they are primarily in the views and we don't have any tests for these at the moment (hence why the `$1` and `$2` weren't picked up).  Maybe we should look into adding request tests, then making the views more consistent?

app/controllers/coronavirus_form/offer_space_type_controller.rb
```diff
@@ -8,34 +8,46 @@ class CoronavirusForm::OfferSpaceTypeController < ApplicationController
...
+  def validate_fields
     [
       validate_field_response_length(controller_name, TEXT_FIELDS),
-      validate_checkbox_field(controller_name, values: offer_space_type, allowed_values: ALLOWED_VALUES),
+      validate_checkbox_field(controller_name, values: @form_responses[:offer_space_type], allowed_values: ALLOWED_VALUES),
       validate_mandatory_text_fields(controller_name, TEXT_FIELDS),
-      validate_description_fields(offer_space_type),
-      validate_descriptions_response_length(offer_space_type),
+      validate_description_fields(@form_responses[:offer_space_type]),
+      validate_descriptions_response_length(@form_responses[:offer_space_type]),
```
leenagupte, line 50: This is something for a future PR, but why are we passing a value to `validate_descriptions_response_length`? I can't see where `_selected_options` are used in the method.

app/controllers/coronavirus_form/product_details_controller.rb
leenagupte: Ah, I guess I was confused because the session is being used to find the product.

app/controllers/coronavirus_form/product_details_controller.rb
```diff
@@ -37,43 +38,45 @@ def destroy
...
-  def validate_missing_fields(product)
+  def validate_missing_fields
     required = []
     required.concat REQUIRED_FIELDS
     required.each_with_object([]) do |field, invalid_fields|
-      next if product[field].present?
+      next if @product[field.to_sym].present?
 
       invalid_fields << {
         field: field.to_s,
```
leenagupte: Not a blocker: Another sneaky `.to_s`. The Copy and Paste is strong with this one.

brucebolt: I'm not sure why `to_s` is used here anyway .... it's already a string 😕

app/views/coronavirus_form/are_you_a_manufacturer.html.erb
leenagupte: Wasn't this supposed to be replaced?

brucebolt: Have actually committed the change this time

leenagupte (commented):
I think one rogue `$1` was left in, but apart from that it looks good. Once that's removed, and I've sanity tested it locally, it will be ready for a ✅

leenagupte (approved):
💯 ⭐️ 👍

PR #201, opened 2020-04-08
Access products with indifferent access

bilbof:
This will prevent issues when using strings or symbols for keys in the product hashes that we store in the session.

https://trello.com/c/5uLlrPfX/184


brucebolt (approved):
👍

PR #204, opened 2020-04-08
Add link to DHSC forms

gclssvglx:
Compromise fix to link to DHSC forms in order to maintain the full CTA.  

Pending a long-term fix.

- Given I've checked 'Testing equipment'
- When I go to 'Tell us more about this product'
- Then I should see a new page with a link to [here](https://www.gov.uk/guidance/help-the-government-increase-coronavirus-covid-19-testing-capacity) - clicking on the "Continue..." button takes you to "Can you offer another product?".

Trello: https://trello.com/c/kvauoDPX/185-add-link-to-dhsc-forms

### DIFF
--- a/app/controllers/coronavirus_form/testing_equipment_controller.rb
+++ b/app/controllers/coronavirus_form/testing_equipment_controller.rb
@@ -1,0 +1,15 @@
+# frozen_string_literal: true
+
+class CoronavirusForm::TestingEquipmentController < ApplicationController
+  def show
+    super
+  end
+
+private
+
+  def previous_path
+    return check_your_answers_url if session["check_answers_seen"]
+
+    medical_equipment_type_url
+  end
+end
--- a/app/views/coronavirus_form/testing_equipment.html.erb
+++ b/app/views/coronavirus_form/testing_equipment.html.erb
@@ -1,0 +1,25 @@
+<% content_for :title do %><%= t('coronavirus_form.testing_equipment.title') %><% end %>
+<% content_for :meta_tags do %>
+  <meta name="description" content="<%= t('coronavirus_form.testing_equipment.title') %>" />
+<% end %>
+
+<% content_for :back_link do %>
+  <%= render "govuk_publishing_components/components/back_link", { href: previous_path } %>
+<% end %>
+
+<%= render "govuk_publishing_components/components/title", {
+  title: t('coronavirus_form.testing_equipment.title'),
+  margin_top: 0
+} %>
+
+<p class="govuk-body govuk-!-margin-bottom-8">
+  <a class="govuk-link" href="<%= t('coronavirus_form.testing_equipment.external_link') %>" target="_blank">
+    <%= t('coronavirus_form.testing_equipment.link.label') %>
+  </a>
+</p>
+
+<%= render "govuk_publishing_components/components/button", {
+  text: t('coronavirus_form.testing_equipment.button.label'),
+  href: additional_product_url,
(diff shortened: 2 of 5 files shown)


app/controllers/coronavirus_form/testing_equipment_controller.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+class CoronavirusForm::TestingEquipmentController < ApplicationController
+  def show
+    session[:product_details] ||= []
+    @product = find_product(params["product_id"], session[:product_details])
+    super
+  end
```
bilbof, line 6: I don't think this method is necessary

app/views/coronavirus_form/testing_equipment.html.erb
```diff
@@ -0,0 +1,26 @@
...
+  <%= render "govuk_publishing_components/components/back_link", { href: previous_path } %>
+<% end %>
+
+<%= render "govuk_publishing_components/components/title", {
+  title: t('coronavirus_form.testing_equipment.title'),
+  margin_top: 0
+} %>
+
+<p class="govuk-body govuk-!-margin-bottom-8">
+  <a class="govuk-link" href="https://www.gov.uk/guidance/help-the-government-increase-coronavirus-covid-19-testing-capacity" target="_blank">
```
bilbof: We could put the href into the `en.yml` file

app/views/coronavirus_form/testing_equipment.html.erb
```diff
@@ -0,0 +1,26 @@
...
+    <%= t('coronavirus_form.testing_equipment.link.label') %>
+  </a>
+</p>
+
+<%= render "govuk_publishing_components/components/button", {
+  text: t('coronavirus_form.testing_equipment.button.label'),
+  href: "/additional-product",
```
bilbof, line 23:
```suggestion
  href: <%= additional_product_url %>,
```

bilbof (approved):
Looks good, a couple of minor changes requested

app/views/coronavirus_form/testing_equipment.html.erb
```diff
@@ -0,0 +1,26 @@
...
+  margin_top: 20,
```
alex-ju: This value is out of range ([since 9 is maximum](https://design-system.service.gov.uk/styles/spacing/)) and we the button component, as most of out components, only supports `margin_bottom` – but I suspect we don't need this line now that we use the `govuk-!-margin-bottom-8` override class on the element above.

app/views/coronavirus_form/testing_equipment.html.erb
```diff
@@ -0,0 +1,26 @@
...
+  margin_bottom: 9
```
alex-ju: I think we should stick with the default value on the button, so the line below should suffice.
```suggestion
  margin_bottom: true
```

alex-ju (commented):
I left two comments frontend-wise.

app/views/coronavirus_form/testing_equipment.html.erb
alex-ju: @bilbof we're already in output mode, I think you're meant to suggest:
```suggestion
  href: additional_product_url,
```

bilbof: Oh yeah you are right, thanks Alex 🤒

bilbof (approved):
⭐️ 🎉

PR #209, opened 2020-04-09
Add email notifications logic

bilbof:
This adds Sidekiq and email notifications logic, using Notify to send emails in a job.

A followup PR will start sending confirmation emails to users when they submit the form, with something like

```ruby
mailer = CoronavirusFormMailer.with(name: session.dig(:contact_details, :contact_name))
mailer.thank_you(session.dig(:contact_details, :email)).deliver_later
```

Co-Authored-By: @GDSNewt 

https://trello.com/c/PFxx7Q7I/148-send-confirmation-email-to-users-upon-completion-business-form


richardTowers:
I think adding a queue (rather then just sending an API call to Notify) is a trade-off we should think about here. 

We get some benefits - it's probably quicker, and we can retry calls when there are errors.

But we also have to maintain a sidekiq queue.

Given that Notify's API is pretty reliable, I wonder if it's worth it?

bilbof:
@richardTowers yeah I considered not using a queue, but I think managing a sidekiq queue won't be much of a burden considering the benefits we get.

In addition to those benefits you've mentioned: user requests will be faster, so we can also handle more of these requests.

Where we've sent emails before in other apps on GOV.UK we've usually ended up adding a queue for handling them later, not only since Notify can be unreliable, but also because of other failures (e.g. bugs in email code).

If we find it is a burden managing the queue, we could stop using Sidekiq in future.

manifest.yml
```diff
@@ -14,5 +14,6 @@ applications:
     GOVUK_APP_DOMAIN: cloudapps.digital
     GOVUK_WEBSITE_ROOT: www.gov.uk
     BASIC_AUTH_USERNAME: gds
+    REDIS_PROVIDER: http://govuk-coronavirus-business-volunteer-form-redis.apps.internal:8080
```
richardTowers: I don't think this URL will work. The redis URL will be provided to the app as part of the VCAP_SERVICES environment variable. `apps.internal` only works for other apps.

richardTowers:
👍  Sounds reasonable, yeah. I'm happy, so long as we've thought it through.

manifest.yml
bilbof: Ah thanks, I wasn't sure about this. I've amended the commit to use VCAP_SERVICES.

brucebolt (commented):
This looks great.  I'm wondering if we're able to define the email content in the locales files?  That way all the content would be in one place, and can be changed by content designers without digging through application code.

bilbof:
@brucebolt good suggestions. I've moved the email subject and body text to the en locales file.

brucebolt (approved):
Looks great!

PR #211, opened 2020-04-09
Fix paths and i18n in feature tests

gclssvglx:
Initially, Feature tests were added with static strings and string type URL's.  

In order to clean up and make the code more maintainable and less brittle, this change converts static strings to I18N strings and string URL's to Rails paths.


bilbof (approved):
Looks great 👍

PR #218, opened 2020-04-11
Run feature tests against staging

gclssvglx:
Run feature tests against staging in the business volunteers pipeline.

Test locally `bundle exec rake`
Test on staging `TEST_URL='https://<USER>:<PASSWORD>@govuk-coronavirus-business-volunteer-form-stg.cloudapps.digital' bundle exec rspec spec/features`

Co-authored with: @bilbof 

Trello: https://trello.com/c/g66yEmsa/44-run-feature-tests-against-staging-in-the-business-volunteers-pipeline

### DIFF
--- a/concourse/pipeline.yml
+++ b/concourse/pipeline.yml
@@ -28,11 +28,33 @@ resources:
       branch: master
       pro: true
 
+  - name: govuk-coronavirus-business-volunteer-form-feature-tests
+    type: git
+    icon: github-circle
+    source:
+      uri: https://github.com/alphagov/govuk-coronavirus-business-volunteer-form
+      branch: master
+      paths:
+        - Gemfile*
+        - .ruby-version
+        - spec/
+
   - name: every-six-hours
     type: time
     icon: clock-outline
     source: {interval: 6h}
 
+  - name: every-day
+    type: time
+    source:
+      interval: 24h
+
+  - name: feature-tests-image
+    type: docker-image
+    icon: docker
+    source:
+      repository: ((readonly_private_ecr_repo_url))
+      tag: govuk-coronavirus-business-volunteer-form-feature-tests
 
 jobs:
   - name: update-pipeline
@@ -41,6 +63,18 @@ jobs:
         trigger: true
       - set_pipeline: govuk-corona-business-volunteer-form
         file: govuk-coronavirus-business-volunteer-form/concourse/pipeline.yml
+
+  - name: build-feature-tests-image
+    serial: true
+    plan:
+      - get: every-day
+        trigger: true
+      - get: govuk-coronavirus-business-volunteer-form-feature-tests
+        trigger: true
+      - put: feature-tests-image
+        params:
+          build: govuk-coronavirus-business-volunteer-form-feature-tests
+          dockerfile: govuk-coronavirus-business-volunteer-form-feature-tests/spec/Dockerfile
 
   - name: deploy-to-staging
     serial: true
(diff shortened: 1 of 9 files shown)


concourse/tasks/run-rspec-test.yml
```diff
@@ -0,0 +1,14 @@
...
+    repository: ((readonly_private_ecr_repo_url))
+    tag: govuk-coronavirus-business-volunteer-form-feature-tests
+params:
+  TEST_URL:
+inputs:
+  - name: govuk-coronavirus-business-volunteer-form
+run:
+  path: rails
+  dir: govuk-coronavirus-business-volunteer-form
+  args: ['rspec']
```
issyl0: If this task is purely for the feature tests, can we run `bundle exec rake spec:features`? It scopes it a bit.

spec/features/filling_in_the_form/fill_in_the_form_spec.rb
```diff
@@ -5,7 +5,7 @@
 RSpec.feature "Fill in the form" do
   include FillInTheFormSteps
 
-  scenario "Complete the form" do
+  scenario "Complete the form", js: true do
```
issyl0: Do we want to have two of these: one with JS and one without?

issyl0 (commented):
This looks great, thanks Graham! Just a couple of thoughts I had. And we can test that it actually works on Concourse next week!

concourse/tasks/run-rspec-test.yml
gclssvglx: Thanks for the review @issyl0 and your comments - updated.

spec/features/filling_in_the_form/fill_in_the_form_spec.rb
gclssvglx: Thanks for the review @issyl0 and your comments - with and without js tests added.

gclssvglx:
> This looks great, thanks Graham! Just a couple of thoughts I had. And we can test that it actually works on Concourse next week!

Thanks @issyl0!  Not 100% sure the syntax is correct on the with and without JS scenarios, but this was the only way I could get it to run and it looks ok if you `bundle exec rspec spec:features --format documentation`.

concourse/pipeline.yml
```diff
@@ -71,6 +77,10 @@ jobs:
         params:
           URL: 'https://govuk-coronavirus-business-volunteer-form-stg.cloudapps.digital/medical-equipment'
           MESSAGE: "Checks that the application deployed to staging is not serving HTTP error codes"
+      - task: run-rspec-specs
```
bilbof: Might be nice to rename this "run smoke tests" since that'll be clearer to folks in future.

spec/features/filling_in_the_form/fill_in_the_form_spec.rb
```diff
@@ -2,46 +2,58 @@
...
   end
 
-  scenario "Ensure we can perform a healthcheck" do
+  describe "fill in the form" do
+    context "without javascript" do
+      it_behaves_like "completing the form"
+    end
+
+    context "with javascript", js: true do
+      it_behaves_like "completing the form"
```
bilbof, line 34: Is this all that's required to run the specs with/without JS? If so, this is a neat way to do it 👍

bilbof (commented):
Looks great! Let's make the change to the concourse pipeline and try it out.

bilbof:
One thing that would be nice to improve before we merge this in is the speed of the test suite. The feature tests take 22 seconds for me locally.

bilbof:
I've rebased and made a couple of tweaks to improve the speed of the tests.

Using apparition as the driver only for the JS tests speeds up the feature tests quite a bit. The JS tests take 10 out of 13.72 seconds from the suite. I wonder if we could mark them as slow / smoke tests with a tag like `@smoke`, so they'd run only in Travis / Concourse.

Let's fix that later.

Another problem is that though JS is enabled, the test isn't catching the bug that is caused when `window.GOVUKFrontend.initAll()` is removed. It would be great if we could figure that out.

bilbof:
Ah so I figured it out. The tests will catch it, but locally one needs to run `rake assets:precompile` to run the JS tests against the latest JS. I think this is acceptable for now, so I'll look at getting this merged and running in concourse.

concourse/pipeline.yml
brucebolt: I think this would be a good idea and would keep this app in line with the vulnerable people form

brucebolt (approved):
Everything looks good to me.  My only comment would be to agree that the `run-rspec-specs` Concourse task should be renamed to be smoke tests.

bilbof:
@brucebolt thanks, I've renamed that.

PR #222, opened 2020-04-14
Remove whitespace from postcodes

leenagupte:
Trello: https://trello.com/c/hUsoCjYe

## Motivation
Users sometimes enter their postcode with extra spaces, e.g "e1 8 qs" which fails validation. This caused quite a few users to contact support.

We need to update the validator so we accept valid postcodes that are entered in the incorrect format, so users are not being rejected when they enter a valid postcode.


bilbof (approved):
Looks good 👍

PR #223, opened 2020-04-14
Fix smoke test task

bilbof:
This would run `rails 'bundle exec rake spec:features'`, but we really want `bundle 'exec rake spec:features'`.

### DIFF
--- a/concourse/tasks/run-smoke-tests.yml
+++ b/concourse/tasks/run-smoke-tests.yml
@@ -9,6 +9,6 @@ params:
 inputs:
   - name: govuk-coronavirus-business-volunteer-form
 run:
-  path: rails
+  path: bundle
   dir: govuk-coronavirus-business-volunteer-form
-  args: ['bundle exec rake spec:features']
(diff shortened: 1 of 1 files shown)


concourse/tasks/run-smoke-tests.yml
```diff
@@ -9,6 +9,6 @@ params:
 inputs:
   - name: govuk-coronavirus-business-volunteer-form
 run:
-  path: rails
+  path: bundle
   dir: govuk-coronavirus-business-volunteer-form
-  args: ['bundle exec rake spec:features']
+  args: ['exec rake spec:features']
```
richardTowers: Don't we want:

```suggestion
  args: ['exec', 'rake', 'spec:features']
```

?

richardTowers (approved):

concourse/tasks/run-smoke-tests.yml
bilbof: Yeah thanks!

PR #224, opened 2020-04-14
Remove Redis from test config

huwd:
Use memory store instead

> If you use :memory_store, the tests should behave in the same way, but be slightly faster and not require redis to be running.


bilbof:
Interesting, thanks Huw! ⭐I think it'd be nice for some tests to use redis, just to check it works with redis. Perhaps we could make this change once we have some integration / system tests that check the DB/redis?

Currently the slowest part of the test suite are the feature tests that require JS, so I don't think we'd see much of a performance gain (yet!).

huwd:
cool, will close for now

PR #225, opened 2020-04-14
Add smoke_test environment

bilbof:
This will get the smoke tests working in concourse.

This new environment will be used for smoke testing only.
E.g. `RAILS_ENV=smoke_test TEST_URL=... bundle exec rake spec:features`

A new environment is required since RSpec requires a DB to be running
and connected to in order to run the tests. We could delete the constant
ActiveRecord e.g. `Object.send(:remove_const, :ActiveRecord)`, but I decided
to use a new env that doesn't load it instead.

I applied this branch to the pipeline and ran a build with it: https://cd.gds-reliability.engineering/teams/govuk-tools/pipelines/govuk-corona-business-volunteer-form/jobs/smoke-test-staging/builds/124

Alternative approaches:

- We could use https://github.com/nulldb/nulldb rather than not requiring
  ActiveRecord
- There might be a way to use rspec/rspec-rails#2266 but I don't think it helps our situations, since we need ActiveRecord for other cases.
- Rewrite the specs using cucumber

https://trello.com/c/g66yEmsa/44-run-feature-tests-against-staging-in-the-business-volunteers-pipeline

### DIFF
--- a/concourse/pipeline.yml
+++ b/concourse/pipeline.yml
@@ -141,7 +141,7 @@ jobs:
       - task: run-smoke-tests
         file: govuk-coronavirus-business-volunteer-form/concourse/tasks/run-smoke-tests.yml
         params:
-          TEST_URL: 'https://coronavirus-business-volunteers.service.gov.uk/medical-equipment'
+          TEST_URL: 'https://coronavirus-business-volunteers.service.gov.uk'
 
   - name: export-form-responses-periodically
     plan:
--- a/concourse/tasks/run-smoke-tests.yml
+++ b/concourse/tasks/run-smoke-tests.yml
@@ -6,9 +6,10 @@ image_resource:
     tag: govuk-coronavirus-business-volunteer-form-feature-tests
 params:
   TEST_URL:
+  RAILS_ENV: smoke_test
 inputs:
   - name: govuk-coronavirus-business-volunteer-form
 run:
   path: bundle
   dir: govuk-coronavirus-business-volunteer-form
-  args: ['exec', 'rake', 'spec:features']
+  args: ['exec', 'rspec', 'spec/features']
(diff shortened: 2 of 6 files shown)


concourse/pipeline.yml
```diff
@@ -141,7 +141,7 @@ jobs:
       - task: run-smoke-tests
         file: govuk-coronavirus-business-volunteer-form/concourse/tasks/run-smoke-tests.yml
         params:
-          TEST_URL: 'https://coronavirus-business-volunteers.service.gov.uk/medical-equipment'
+          TEST_URL: 'https://coronavirus-business-volunteers.service.gov.uk'
```
issyl0, line 144: This should go back to the way it was. The `/` route redirects back to GOV.UK, and [we don't have any feature tests that click "Start now"](https://github.com/alphagov/govuk-coronavirus-business-volunteer-form/blob/f334af7ffe7ebc29faffe444ef1dab4d38588f53/spec/support/fill_in_the_form_steps.rb#L4-L6). Probably want to remove the first feature, actually...

bilbof: Ah sorry I should've separated this out into another commit for clarity. This change is intended, since `TEST_URL` is used by the feature tests as the `Capybara.app_host`, so when one writes `visit "medical_equipment_path"` behind the scenes this gets translated into `Capybara.app_host + path` or `https://coronavirus-business-volunteers.service.gov.uk/medical-equipment`.

The previous commit was broken, so if I left it as-is the calls to `visit_path` would take the test browser to e.g. `/medical-equipment/medical-equipment/`.

issyl0: Ah, that makes more sense!

issyl0 (approved):

PR #227, opened 2020-04-15
Run tests every time a change is pushed

leenagupte:
This should run tests every time a change is pushed to any branch as well as pull requests


issyl0 (approved):
Thanks!

PR #229, opened 2020-04-15
Fix asset sync pipeline

bilbof:
This environment variable was set on the wrong app, introduced in https://github.com/alphagov/govuk-coronavirus-business-volunteer-form/pull/187


issyl0 (approved):
Oops.

PR #231, opened 2020-04-15
Add extra validation for product details

leenagupte:
Trello: https://trello.com/c/hjNli4xS

# What

Add validation to Product Price, Product Quantity and Product lead time on the Business Volunteering form.

Requirements:

**General validation notes:**

- Strip whitespace from the start and end of fields before validating

**These all apply to the product details page:**

Quantity

- Strip commas
- Validate for a positive integer
- Validation error message: "Enter the approximate total number of items, do not include words or symbols"

Cost

- Strip £ and 'GBP' (case insensitive)
- Validate for a positive integer, optionally with 2 decimal places
- Validation error message: "Enter the cost per item in pounds, do not include words or symbols. If you are giving it for free, enter 0"

Lead time

- Strip 'days' (case insensitive)
- Validate for a positive integer
- Validation error message: "Enter the approximate lead time in days, do not include words or symbols. If the product is available now, enter 0"

# Why

Cabinet Office told us they couldn't use these fields, since folks are treating these fields as free text rather than numbers. E.g. users would enter 'about 3 weeks' for the lead time in days field.

**Details of fields**

Volumes:
- Need in order to prioritise the buying of PPE/Testing and other medical equipment
- Need for daily reporting numbers to number 10

Values:
- Need for daily reporting to number 10

Lead times:
- Needed to co-ordinate logistics to get PPE/Medical equipment to the NHS
- Needed to prioritise the buying of PPE/Testing and other medical equipment
- Needed for daily reporting to number 10

# Expected changes
## Product quantity
quantity_error

## Product cost
cost_error

## Lead time
lead_time_error

### DIFF
--- a/app/controllers/coronavirus_form/product_details_controller.rb
+++ b/app/controllers/coronavirus_form/product_details_controller.rb
@@ -81,55 +81,44 @@ private
     required.each_with_object([]) do |field, invalid_fields|
       next if @product[field.to_sym].present?
 
-      invalid_fields << {
-        field: field,
-        text: t("coronavirus_form.questions.#{controller_name}.#{field}.custom_error",
-                default: t("coronavirus_form.errors.missing_mandatory_text_field",
-                           field: t("coronavirus_form.questions.#{controller_name}.#{field}.label")).humanize),
-      }
+      invalid_fields << error_response(field)
     end
   end
 
   def validate_product_quantity
-    quantity = @product[:product_quantity]
-    error_response = {
-      field: "product_quantity",
-      text: t("coronavirus_form.questions.#{controller_name}.product_quantity.custom_error"),
-    }
+    field = "product_quantity"
+    return error_response(field) unless @product[field.to_sym]
 
-    return if Integer(quantity) && Integer(quantity).positive?
+    quantity = Integer(@product[field.to_sym])
+    return if quantity && quantity.positive?
 
-    error_response
-  rescue TypeError, ArgumentError
-    error_response
+    error_response(field)
+  rescue ArgumentError
+    error_response(field)
   end
 
   def validate_product_cost
-    cost = @product[:product_cost]
-    error_response = {
-      field: "product_cost",
-      text: t("coronavirus_form.questions.#{controller_name}.product_cost.custom_error"),
-    }
+    field = "product_cost"
+    return error_response(field) unless @product[field.to_sym]
 
-    return if Float(cost) && (Float(cost).positive? || Float(cost).zero?)
+    cost = Float(@product[field.to_sym])
+    return if cost && (cost.positive? || cost.zero?)
 
-    error_response
-  rescue TypeError, ArgumentError
-    error_response
+    error_response(field)
+  rescue ArgumentError
+    error_response(field)
   end
 
   def validate_lead_time
-    days = @product[:lead_time]
-    error_response = {
-      field: "lead_time",
-      text: t("coronavirus_form.questions.#{controller_name}.lead_time.custom_error"),
-    }
+    field = "lead_time"
+    return error_response(field) unless @product[field.to_sym]
 
-    return if Integer(days) && (Integer(days).positive? || Integer(days).zero?)
+    days = Integer(@product[field.to_sym])
+    return if days && (days.positive? || days.zero?)
 
-    error_response
-  rescue TypeError, ArgumentError
-    error_response
+    error_response(field)
+  rescue ArgumentError
+    error_response(field)
   end
 
   def sanitized_product(params)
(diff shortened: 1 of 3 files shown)


app/controllers/coronavirus_form/product_details_controller.rb
```diff
@@ -78,27 +81,61 @@ def validate_missing_fields
...
     end
   end
 
+  def validate_product_quantity
+    field = "product_quantity"
+    quantity = Integer(@product[field.to_sym])
+    return if quantity && quantity.positive?
+
+    error_response(field)
+  rescue TypeError, ArgumentError
```
bilbof: This is reasonable 👍I wonder if it might hide genuine bugs from Sentry. Would it be possible to guard against these errors with e.g. `return error_response(field) unless @product[field.to_sym]`?

app/controllers/coronavirus_form/product_details_controller.rb
```diff
@@ -78,27 +81,61 @@ def validate_missing_fields
...
+    error_response(field)
+  end
+
+  def validate_product_cost
+    field = "product_cost"
+    cost = Float(@product[field.to_sym])
+    return if cost && (cost.positive? || cost.zero?) && @product[field.to_sym].match?(/^\d+\.?\d{0,2}$/)
```
bilbof: Would it be possible to create a constant for the regex? E.g. `VALID_FLOAT = /^\d+\.?\d{0,2}$/`, I usually prefer this since one can see at a glance what the intent of the regex is.

app/controllers/coronavirus_form/product_details_controller.rb
```diff
@@ -78,27 +81,61 @@ def validate_missing_fields
...
+
   def sanitized_product(params)
     {
       product_id: strip_tags(params[:product_id]).presence || SecureRandom.uuid,
       equipment_type: strip_tags(params[:equipment_type]).presence,
       product_name: strip_tags(params[:product_name]).presence,
-      product_quantity: strip_tags(params[:product_quantity]).presence,
-      product_cost: strip_tags(params[:product_cost]).presence,
+      product_quantity: strip_tags(params[:product_quantity]&.gsub(",", "")&.strip).presence,
+      product_cost: strip_tags(params[:product_cost]&.gsub(/[,£a-zA-Z]/, "")&.strip).presence,
```
bilbof, line 131: If a user enters "About £4.00 for a box", we'll use `4.00`. I wonder if this will be confusing later on when the case manager asks the user about product price, if the number the user enters doesn't relate to single units? Though, this might not be something we can solve through validation.

spec/controllers/coronavirus_form/product_details_controller_spec.rb
```diff
@@ -203,6 +203,99 @@
...
+        before do
+          params.merge!(equipment_type: "Gloves")
+        end
+
+        it "errors if the user doesn't provide a product_quantity" do
+          post :submit, params: params.except(:product_quantity)
+          expect(response).to render_template(current_template)
+        end
+
+        it "errors if the user enters and invalid product quantity" do
```
bilbof: This should be "an".

spec/controllers/coronavirus_form/product_details_controller_spec.rb
```diff
@@ -203,6 +203,99 @@
...
+        end
+
+        it "allows the lead_time to be zero" do
+          post :submit, params: params.merge(lead_time: "0")
+          expect(response).to redirect_to(additional_product_path)
+        end
+
+        it "removes words from lead_time" do
+          post :submit, params: params.merge(lead_time: "10 days")
+          expect(session[:product_details].first[:lead_time]).to eq("10")
```
bilbof, line 296: Hm, this might be confusing is someone writes "10 weeks" into the field. Do you think we should return an error instead?

bilbof (changes requested):
Looks good, just some general questions

app/controllers/coronavirus_form/product_details_controller.rb
leenagupte: Yeah, the validation is really to help on the data-side of things. I mean, how big is a box?

spec/controllers/coronavirus_form/product_details_controller_spec.rb
leenagupte: The requested validation was just to strip out "days". But then I thought, what if the user enters "1 day", so I stripped out all words. Now I'm not sure which approach to take.

app/controllers/coronavirus_form/product_details_controller.rb
leenagupte: 👍 Doing that would remove the need to catch the TypeError.

leenagupte:
@bilbof Thanks for your review. I've made the suggested changes and fixed up, and responded to your comments.

app/controllers/coronavirus_form/product_details_controller.rb
```diff
@@ -78,27 +82,67 @@ def validate_missing_fields
...
 
+  def validate_product_quantity
+    field = "product_quantity"
+    return error_response(field) unless @product[field.to_sym]
+
+    quantity = Integer(@product[field.to_sym])
+    return if quantity && quantity.positive?
+
+    error_response(field)
+  rescue ArgumentError
```
bilbof, line 97: Looks better, thanks 👍

Perhaps we could do a `.gsub(/[a-zA-Z]| /, "")` to try and prevent argument errors, but it might bring up the same issues as mentioned before. Attempting to correct the user might confuse things when they get asked about their "10100" items but they wrote "between 10 and 100".

We could also try using `.to_i` here, which gives `0` as a fallback, rather than raising an error. Would want to wrap all this logic in a kind of `UserProvidedInteger` class, but creating all this feels a bit like building GOV.UK Submit :)

I thought way too long about it - this seems good to me 👍

bilbof (approved):
Looks good. Sorry for confusing things with my questions. This will definitely help those consuming the data downstream 👍 

Perhaps we should keep an eye on 422 / conversion rates when we deploy this to see if more users get stuck.

PR #240, opened 2020-04-16
Send staging emails to courtesy copy group

bilbof:
https://trello.com/c/U4Xn4RrU/259-send-staging-emails-to-courtesy-copy-group


app/controllers/coronavirus_form/check_answers_controller.rb
```diff
@@ -22,6 +21,19 @@ def submit
 
 private
 
+  def thank_user
```
brucebolt: I'd suggest changing this method name to something like `send_confirmation_email` since I was initially expecting it to render a thank you view 😁

brucebolt (approved):
Just left a comment about renaming a method, otherwise this looks great.

PR #243, opened 2020-04-17
Update concourse

huwd:
Trello: https://trello.com/c/t6jxzyCY/229-dockerize-app-and-run-tests-within-concourse

## What
- Adds a single docker image we can use for all test builds

- Give resources shorter, more generic names. It is clear from within a
pipeline which service we are in, and long names make for things
bunching up on the UI, so generally simplify everything down to just
`git-master` or simple descriptions of what the resource actually is.

- Create a run quality check task, this should be able to take a
concourse test docker image with all the system level dependencies we
need (Redis/Postgres) and then provide all the commands needed to set up
and run the test suite + linting.

- Rearrange the triggers in the pipeline so quality checks run every
time we get a change to the master branch. These quality checks will
then be a dependency for deployments to staging and production. If they
fail, then we don't go any further.

## Why

Allows us great reassurance that any build heading down a pipeline to production passes our required quality checks.

## Follow up
After this, I'll put in a separate tidy up PR that will:
- Remove all travis config
- Remove the features Dockerfile
- Tidy up anything else we no longer need after this.

### DIFF
--- a/concourse/Dockerfile
+++ b/concourse/Dockerfile
@@ -1,0 +1,33 @@
+#========================================================================================
+# This Dockerfile creates a container image which contains Ruby,
+# Google Chrome and Postgres, as well as the application source and bundled dependencies.
+#
+# This is needed for running the feature tests in a container environment like Concourse.
+#========================================================================================
+FROM ruby:2.6.5-buster
+
+# Updates all debian dependencies
+# Adds google chrome to software sources, install chrome, install postgres
+# clear up lists behind ourselves
+RUN apt-get update --fix-missing && apt-get -y upgrade \
+    && wget -q -O - https://dl-ssl.google.com/linux/linux_signing_key.pub | apt-key add - \
+    && sh -c 'echo "deb [arch=amd64] http://dl.google.com/linux/chrome/deb/ stable main" >> /etc/apt/sources.list.d/google.list' \
+    && apt-get update \
+    && apt-get install -y --no-install-recommends \
+      google-chrome-unstable \
+      postgresql-11 \
+      redis-server \
+    && rm -rf /var/lib/apt/lists/* \
+    && rm -rf /src/*.deb
+
+ENV CHROME_NO_SANDBOX=true
+
+COPY Gemfile* .ruby-version /application/
+
+WORKDIR /application/
+
+RUN bundle install
+
+COPY . /application/
+
+ENTRYPOINT ["/bin/bash"]
--- a/concourse/Dockerfile
+++ b/concourse/Dockerfile
@@ -1,13 +1,13 @@
 #========================================================================================
 # This Dockerfile creates a container image which contains Ruby,
-# Google Chrome and Postgres, as well as the application source and bundled dependencies.
+# Google Chrome, Redis and Postgres, as well as the application source and bundled dependencies.
 #
 # This is needed for running the feature tests in a container environment like Concourse.
 #========================================================================================
 FROM ruby:2.6.5-buster
 
 # Updates all debian dependencies
-# Adds google chrome to software sources, install chrome, install postgres
+# Adds google chrome to software sources, install chrome, install postgres and redis
 # clear up lists behind ourselves
 RUN apt-get update --fix-missing && apt-get -y upgrade \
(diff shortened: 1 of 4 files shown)


concourse/pipeline.yml
```diff
@@ -5,39 +5,23 @@ resource_types:
...
-      pro: true
-
-  - name: govuk-coronavirus-business-volunteer-form-feature-tests
-    type: git
+  - name: git-master-gems-dockerfile
     icon: github-circle
+    type: git
     source:
-      uri: https://github.com/alphagov/govuk-coronavirus-business-volunteer-form
+      uri:  https://github.com/alphagov/govuk-coronavirus-business-volunteer-form
```
issyl0:
```suggestion
      uri: https://github.com/alphagov/govuk-coronavirus-business-volunteer-form
```

concourse/Dockerfile
```diff
@@ -0,0 +1,33 @@
+#========================================================================================
+# This Dockerfile creates a container image which contains Ruby,
+# Google Chrome and Postgres, as well as the application source and bundled dependencies.
```
huwd:
```suggestion
# Google Chrome, Redis and Postgres, as well as the application source and bundled dependencies.
```

concourse/Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+#
+# This is needed for running the feature tests in a container environment like Concourse.
+#========================================================================================
+FROM ruby:2.6.5-buster
+
+# Updates all debian dependencies
+# Adds google chrome to software sources, install chrome, install postgres
```
huwd:
```suggestion
# Adds google chrome to software sources, install chrome, install postgres and redis
```

concourse/Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+# clear up lists behind ourselves
+RUN apt-get update --fix-missing && apt-get -y upgrade \
+    && wget -q -O - https://dl-ssl.google.com/linux/linux_signing_key.pub | apt-key add - \
```
huwd, line 13: Does this app still need chrome headless?
Rspec feature tests use it right?

concourse/pipeline.yml
huwd:
```suggestion
      uri: https://github.com/alphagov/govuk-coronavirus-business-volunteer-form
```

concourse/pipeline.yml
```diff
@@ -5,39 +5,23 @@ resource_types:
...
       branch: master
       paths:
+        - concourse/Dockerfile
         - Gemfile*
-        - .ruby-version
-        - spec/
```
huwd, line 40: Not sure we need to have conrouse monitor the `/spec` file for this, we'd only want to rebuild this for dependency or docker changes i think :thinking:

concourse/pipeline.yml
```diff
@@ -49,42 +33,77 @@ resources:
...
+            dir: git-master
+            path: bash
+            args:
+              - -c
+              - |
+                set -eu
+                service postgresql start
+                su - postgres -c "psql -c \"create role root with createdb login password 'password';\""
+                service redis-server start
+                export TEST_DATABASE_URL="postgres://root:password@localhost:5432/coronavirus_business_volunteer_form_test"
```
huwd, line 91: Should reflect this stuff back on the find-support form.
I think the app thinks it's running in development over there.

huwd:
Interesting Heroku errored, is that happening elsewhere?

concourse/pipeline.yml
issyl0: If the Dockerfile to build the features image is in `concourse/` now then no, that's fine, we don't need `spec`. We probably _do_ want when `.ruby-version` changes, but with that would come a `Dockerfile` change to pull `ruby:2.6.6-buster` so I think that's fine to leave out.

huwd: :+1:

concourse/Dockerfile
issyl0: I believe that's right.

issyl0 (approved):

PR #244, opened 2020-04-17
Remove travis

huwd:
# What
Removes travis
Removes a spec docker file now replaced by concourse/dockerfile


issyl0 (approved):
Feels gooood. Good work! 💪

PR #249, opened 2020-04-20
Prevent users submitting the same FormResponse twice

bilbof:
Currently a user can submit the form twice if they click the submit button on the check your answers page twice in quite succession.

This adds two fixes for this:

- The frontend fix is to disable the button with javascript when the form is submitted; this prevents users with javascript enabled clicking the submit button multiple times
- The backend fix uses the unique constraint we have on FormResponse data; it sets reference_id on show, before submit is clicked, so the reference_id doesn't change when users click 'submit' multiple times. This means that subsequent requests will return an error, rather than creating duplicate records in the DB.

https://trello.com/c/cQxG2uZz/284

### DIFF
--- a/app/controllers/coronavirus_form/check_answers_controller.rb
+++ b/app/controllers/coronavirus_form/check_answers_controller.rb
@@ -1,25 +1,29 @@
 # frozen_string_literal: true
 
 class CoronavirusForm::CheckAnswersController < ApplicationController
+  before_action :set_reference_number
+
   def show
     session[:check_answers_seen] = true
     super
   end
 
   def submit
-    submission_reference = reference_number
-
-    session[:reference_number] = submission_reference
     FormResponse.create(form_response: session) unless smoke_tester?
 
     send_confirmation_email
 
+    ref_number = session[:reference_number]
     reset_session
 
-    redirect_to thank_you_url(reference_number: submission_reference)
+    redirect_to thank_you_url(reference_number: ref_number)
   end
 
 private
+
+  def set_reference_number
+    session[:reference_number] ||= reference_number
+  end
 
   def send_confirmation_email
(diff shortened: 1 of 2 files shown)


alex-ju (commented):
We already have this functionality available (admittedly not documented in the components gem yet, as it's a feature coming from the Design System), we just need to turn it on using the data attributes as follows:
```
<%= render "govuk_publishing_components/components/button", {
  text: t("coronavirus_form.submit_and_next"),
  margin_bottom: true,
  data_attributes: {
    "module": "govuk-button",
    "prevent-double-click": "true"
  }
} %>
```

bilbof:
@alex-ju ah awesome, I should've asked!

I tried this locally and it didn't seem to disable the button / form - users could still submit the form multiple times. Is the functionality ready for use?

alex-ju:
> I tried this locally and it didn't seem to disable the button / form - users could still submit the form multiple times. Is the functionality ready for use?

It doesn't disable the button visually, it's [de-bouncing the click event when double-clicked (clicked multiple times within a second)](https://github.com/alphagov/govuk-frontend/blob/master/src/govuk/components/button/button.js#L50-L52).

bilbof:
Ah I see. When trying it locally I was able to submit the form multiple times, I guess since valid requests can take longer than a second.

I prefer disabling the button, since it gives feedback that the request is working and prevents re-submitting, but if the request fails I guess we'd want to re-enable the button to allow for re-submits 🤔

alex-ju:
> but if the request fails I guess we'd want to re-enable the button to allow for re-submits

Yeah, that's my only concern with the current approach.

bilbof:
@alex-ju cool, I guess this logic should be done in the components anyway, so I've switched to using the option you recommended. If we want improved behaviour here we could make that change in the component later.

alex-ju (approved):
👍

PR #252, opened 2020-04-20
Increase timeout for page loads in feature tests

bilbof:
This is to avoid any flakiness of feature tests related to slow page loads.

The default is 30 seconds, so this reverts it back to 30.


sihugh (approved):
Seems long, but OK!

bilbof:
Yeah it's a bit long! I've seen the odd test run where a feature test had been failed for no reason, so I'm hoping this'll make those mysteriously go away.

PR #254, opened 2020-04-21
Add ability to return 422 for validation errors

GDSNewt:
This will help us see where users are hitting validation errors, and if we're seeing elevated numbers of errors. This'll help us debug and investigate issues. 

https://trello.com/c/n31BWiHm/203-return-a-422-for-validation-errors-in-business-form


brucebolt (approved):
Looks good 👍.

Could you add a link to the Trello card in the PR description?  It'll then appear in the Trello card automatically, which makes it easier to track which code change belongs to specific pieces of work.

GDSNewt:
> Looks good 👍.
> 
> Could you add a link to the Trello card in the PR description? It'll then appear in the Trello card automatically, which makes it easier to track which code change belongs to specific pieces of work.

Thanks Bruce, I think I was adding it as you wrote this 😄

spec/controllers/coronavirus_form/business_details_controller_spec.rb
```diff
@@ -83,8 +83,7 @@
         company_location: I18n.t("coronavirus_form.questions.business_details.company_location.options.european_union.label"),
       ).except(:company_postcode)
 
-      expect(response).to have_http_status(:unprocessable_entity)
-      expect(response).to render_template(current_template)
+      expect(response).to redirect_to(contact_details_path)
```
brucebolt: This merge commit seems to have removed the `expect(response).to have_http_status(:unprocessable_entity)` that you made in a previous commit

GDSNewt: Hi Bruce, that was intentional. I thought if we were expecting a redirect then we should not be expecting a 422 error. At least i think that's what it's supposed to be doing 🤔

PR #255, opened 2020-04-21
Fix business details specs

bilbof:
These specs weren't working correctly as they need to
have default params that would result in a successful
response.


gclssvglx (approved):
LGTM 👍

PR #270, opened 2020-04-27
Schema expectation tests

leenagupte:
Trello: https://trello.com/c/FEZ30Fa2

## What's changed?

Creates a JSON schema for the FormSubmission model and adds tests to ensure that user form submissions are valid against the schema.

## Why

Currently we need to email Cabinet Office and their contractors when we make a change to the form. This isn't a sustainable process, so instead we'd like an automated test to ensure that we don't break things for folks downstream who are using the data we're generating.

## How to test locally

1. Run the tests locally
2. Run the app locally and submit the form. The form should submit successfully. 
3. There should be no changes to the user's experience.


app/models/form_response.rb
```diff
@@ -1,4 +1,7 @@
 # frozen_string_literal: true
 
 class FormResponse < ApplicationRecord
+  JSON_SCHEMA = Rails.root.join("config/schemas/form_response.json").to_s
+
+  validates :form_response_v2, presence: true, json: { message: ->(errors) { errors }, schema: JSON_SCHEMA }
```
bilbof: I wonder if we can find a way to log validation errors here, but not prevent users from submitting data, until we're sure the schema is complete, to be sure we'll not prevent users from submitting valid data.

config/schemas/form_response.json
```diff
@@ -0,0 +1,329 @@
...
+    "postcode": {
+      "type": ["string", "null"],
+      "pattern": "^[A-Za-z0-9 ]{4,}$"
+    },
+    "whole_number": {
+      "type": "string",
+      "pattern": "^[0-9]+$"
+    }
+  }
+}
```
bilbof, line 329: This looks great 🎉 

Since we might change the translations file without updating this one, that might cause some interesting issues.

One thing that might be interesting is to create a `form_response.json.erb` that we can use to generate a valid schema from `en.yml` changes. That might prevent us from shipping a change that allows users to submit invalid data and highlight in pull requests that the schema is changing, so we can version it / let stakeholders know. What do you think?

bilbof (commented):
Looks great 👍

app/models/form_response.rb
leenagupte: I don't think that's how validations are designed to work. I'm using `create!` rather than `create` so an exception is raised, which I've assumed will be logged in Sentry.

I think this would work though in the controller though:

```
fr = FormResponse.new(form_response: session)
if fr.invalid? 
  logger.info do
      {
        validation_error: { text: "schema validation failed" },
      }.to_json
    end
end

fr.save(validate: false)
```

I'd need to find away to get the actual error message, otherwise the log entry will be useless.

leenagupte: Or

```
    unless smoke_tester?
      begin
        FormResponse.create!(form_response: session.to_hash)
      rescue ActiveRecord::RecordInvalid => e
        logger.info do
          {
            validation_error: { text: e.message },
          }.to_json

        FormResponse.create(form_response: session.to_hash, validate: false)
      end
    end
```

config/schemas/form_response.json
bilbof: We discussed offline and decided we'll do this in another PR (might need to do a bit of cleanup of the en.yml file).

huwd:
Humm, why'd the branch not get deployed to heroku?
image

huwd:
Code looks great! Congrats, will clone it down and run those tests first thing tomorrow.

app/models/form_response.rb
bilbof: Sorry forgot to add a comment on this one. We discussed offline - iirc the plan is to send an error to sentry if the data isn't valid against the schema, but not prevent users submitting the form, and keeping the existing model validations in place if possible.

app/controllers/coronavirus_form/check_answers_controller.rb
```diff
@@ -9,7 +9,16 @@ def show
...
+    unless smoke_tester?
+      form_response = FormResponse.new(form_response: session.to_hash)
+
+      begin
+        form_response.save!
+      rescue ActiveRecord::RecordInvalid => e
+        GovukError.notify(e.message)
+        form_response.save(validate: false)
+      end
+    end
```
bilbof, line 20: This will work 👍Though, would this currently permit invalid data for any field?

Since this is to ensure we don't prevent users from submitting valid data, perhaps we could check that the form_response is valid against the JSON schema outside of an active record validation.

This could happen in the controller, or perhaps in a validation helper that always passes but sends an error to Sentry if the data doesn't pass the schema.

For example, search-api tests that [a content item is valid against a schema](https://github.com/alphagov/search-api/blob/4dd65e29239d8b3f878ffc93cabd76b10e0de8bd/spec/unit/content_item_publisher/finder_presenter_spec.rb#L25). This is done in a test, but I reckon we could do the same assertion in a controller, e.g.

```ruby
# in check your answers controller, before FormResponse.new(...).save!
unless valid_against_schema(form_response)
  GovukError.notify("invalid")
end
```

Later, once we're happy we're not going to stop people from submitting valid responses, we could introduce an active record validation for this data. What do you think?

app/controllers/coronavirus_form/check_answers_controller.rb
```diff
@@ -9,7 +9,16 @@ def show
...
 
   def submit
-    FormResponse.create(form_response: session) unless smoke_tester?
+    unless smoke_tester?
+      form_response = FormResponse.new(form_response: session.to_hash)
+
+      begin
+        form_response.save!
+      rescue ActiveRecord::RecordInvalid => e
+        GovukError.notify(e.message)
```
bilbof: I think this could be `GovukError.notify(e)`

bilbof (changes requested):
This looks good, nice work! I added a couple comments.

Could you also provide a before and after of what a row of the data export will look like? We can send this to our stakeholders at Cabinet Office to give them a heads up of the change.

huwd:
Finally got my ruby sorted, have cloned this down, run the tests and checked it can resubmit. All great.
So looks good to me!

Guess there's just Bills points?

huwd (dismissed):

app/controllers/coronavirus_form/check_answers_controller.rb
leenagupte: Sure! I've made the changes and fixed up. As the validation has been separated from the model, I don't have to change how the session is stored, that means there's no risk of any changes to the data export.

leenagupte:
@bilbof @huwd Thanks for you reviews. I've removed the ActiveRecord, and am now just validating the json before the form is submitted. This means there are no changes to how the data is being stored. The changes are ready for a re-review.

app/controllers/coronavirus_form/check_answers_controller.rb
```diff
@@ -9,6 +11,9 @@ def show
   end
 
   def submit
+    validation_errors = validate_against_form_response_schema(session.to_h)
+    GovukError.notify(validation_errors) if validation_errors.any?
```
bilbof: This might want an exception or a string, I'm not sure what an object will do to it - it might work

https://github.com/alphagov/govuk_app_config/blob/master/lib/govuk_app_config/govuk_error.rb#L5

bilbof (approved):
Excellent work 👏

I had a question about what we'll be sending to `GovukError.notify`, but if that's intended lets 🚢

app/controllers/coronavirus_form/check_answers_controller.rb
leenagupte: I could `.to_s` the array?

bilbof: It would be nice to keep the structure... perhaps you could have that as the `extra: ...` part e.g. `GovukError.notify(FormResponseInvalid.new, extra: ...`), and provide a new error class? I'm guessing that would help Sentry group all of these together under one issue.

leenagupte: I'll do that and fix up.

bilbof (approved):
🎉

leenagupte:
@bilbof I've created a custom error to send to Sentry and fixed up.

bilbof (approved):
Awesome. Great work! 🚢

bilbof (approved):
Awesome. Great work! 🚢

PR #275, opened 2020-05-04
Add suffix to product detail inputs

GDSNewt:
What
----

Add suffixes to the product_quantity and lead_time fields (See attached screenshot).

Cabinet Office told us they couldn't use these fields, as users were treating these fields as free text rather than numbers. E.g. users would enter 'about 3 weeks' for the lead time in days field.

The intention behind the suffixes is to better indicate to users how we would like them to input data.

No suffix was added to the product cost field as desired in the [initial design](https://trello-attachments.s3.amazonaws.com/5e8346012feac37ef24b2a9e/5e983fb16ed1383ba60e8ad4/1ec551844ac2fce3eb0b2c0a4ca68c31/suffixes.png) for this ticket, as it had been decided that a prefix is more appropriate in this instance. We are developing that as part of this future [ticket ](https://trello.com/c/GCGNaCSH/367-add-prefix-to-input-fields-on-product-details-question). 

**BEFORE**
Screenshot 2020-05-06 at 10 39 10

**AFTER**
Screenshot 2020-05-06 at 10 37 57

Links
-----

[Trello Card](https://trello.com/c/7VVBYT7z/257-add-suffix-to-input-fields-on-product-details-question)

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -227,6 +227,7 @@ en:
           label: "Quantity of this product"
           hint: "Approximate amount, for example 10, 500, 10000."
           custom_error: "Enter the approximate total number of items, do not include words or symbols"
+          suffix: "items"
         product_cost:
           label: "Cost per item, in pounds"
           hint: "For example, 23.99. Enter 0 if you will donate it."
@@ -253,6 +254,7 @@ en:
         lead_time:
           label: "Lead time in days"
           custom_error: "Enter the approximate lead time in days, do not include words or symbols. If the product is available now, enter 0"
+          suffix: "days"
         equipment_type:
           label: "What type of equipment is it?"
           options:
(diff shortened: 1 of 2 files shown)


config/locales/en.yml
```diff
@@ -227,6 +227,7 @@ en:
           label: "Quantity of this product"
           hint: "Approximate amount, for example 10, 500, 10000."
           custom_error: "Enter the approximate total number of items, do not include words or symbols"
+          suffix: "Items"
```
brucebolt:
```suggestion
          suffix: "items"
```

config/locales/en.yml
```diff
@@ -253,6 +254,7 @@ en:
         lead_time:
           label: "Lead time in days"
           custom_error: "Enter the approximate lead time in days, do not include words or symbols. If the product is available now, enter 0"
+          suffix: "Days"
```
brucebolt:
```suggestion
          suffix: "days"
```

brucebolt (changes requested):
Looks good, but the screenshot in the Trello card has the suffixes with a lowercase first letter.

config/locales/en.yml
GDSNewt: Thanks Bruce, i've changed them both to be lowercase

brucebolt (approved):
🎉

PR #277, opened 2020-05-04
Make offer_other_support optional in the schema

leenagupte:
Users don't have to answer this question as they can click continue without providing an answer in the form,
so we should also make this question optional in the schema.

Fixes:

https://sentry.io/organizations/govuk/issues/1639294230/events/80be6cc2ce454a15b1e040c26c935273/?project=5172100&query=is%3Aunresolved


bilbof (approved):
🎉

PR #281, opened 2020-05-05
Fix error validations for /offer-care-qualifications

Rosa-Fox:
What
----

- Previously the 'Select if you can offer care for adults or children' error message incorrectly also displayed in the 'What qualifications or certificates do you have?' section
- It was possible to select a qualification and 'I do not have a qualification'

Before screenshot:

before-dont-select-offer-care

Describe what you have changed and why.

- The correct error message now displays in the 'What qualification or certificates do you have?' section
- New validation added to make it impossible to select a qualification and 'I do not have a qualification'

After screen shots:

after-dont-select-offer-care

after-qualifications-section-not-selected

after-qualification-and-no-qualification-selected

Links
-----
[Trello](https://trello.com/c/Hch0J2Zp/272-fix-error-summary-on-what-kind-of-care-can-you-offer)

### DIFF
--- a/app/controllers/coronavirus_form/offer_care_qualifications_controller.rb
+++ b/app/controllers/coronavirus_form/offer_care_qualifications_controller.rb
@@ -10,19 +10,7 @@ class CoronavirusForm::OfferCareQualificationsController < ApplicationController
       offer_care_qualifications_type: strip_tags(params[:offer_care_qualifications_type]).presence,
     }
 
-    invalid_fields = validate_field_response_length("#{controller_name}.care_qualifcations", TEXT_FIELDS) +
-      validate_checkbox_field(
-        "#{controller_name}.offer_care_type",
-        values: @form_responses[:offer_care_type],
-        allowed_values: I18n.t("coronavirus_form.questions.#{controller_name}.offer_care_type.options").map { |_, item| item.dig(:label) },
-      ) +
-      validate_checkbox_field(
-        "#{controller_name}.care_qualifications",
-        values: @form_responses[:offer_care_qualifications],
-        allowed_values: I18n.t("coronavirus_form.questions.#{controller_name}.care_qualifications.options").map { |_, item| item.dig(:label) },
-        other: @form_responses[:offer_care_qualifications_type],
-        other_field: "nursing_or_healthcare_qualification",
-      )
+    invalid_fields = validate_fields
 
     if invalid_fields.any?
       flash.now[:validation] = invalid_fields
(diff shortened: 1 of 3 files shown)


app/controllers/coronavirus_form/offer_care_qualifications_controller.rb
```diff
@@ -22,7 +22,8 @@ def submit
         allowed_values: I18n.t("coronavirus_form.questions.#{controller_name}.care_qualifications.options").map { |_, item| item.dig(:label) },
         other: @form_responses[:offer_care_qualifications_type],
         other_field: "nursing_or_healthcare_qualification",
-      )
+      ) +
```
brucebolt: Since there are now quite a few validations, I'm wondering whether the syntax we've used in other controllers (where we have an array that is flattened) may be slightly more elegant?  E.g. https://github.com/alphagov/govuk-coronavirus-business-volunteer-form/blob/0bc813c0afd1e0ebb3f1bd58e3d67e6fe969c644/app/controllers/coronavirus_form/contact_details_controller.rb#L31-L37

config/locales/en.yml
```diff
@@ -443,6 +443,7 @@ en:
               label: "I do not have a qualification"
           or: "Or"
           custom_select_error: "Select the qualifications that you or people in your company have. If you do not have any select ‘I do not have a qualification’"
+          custom_select_error_qualification_or_not: "You cannot select a qualification as well as 'I do not have a qualification'"
```
brucebolt:
```suggestion
          custom_select_error_qualification_or_not: "You cannot select a qualification as well as ’I do not have a qualification’"
```

app/helpers/field_validation_helper.rb
```diff
@@ -74,6 +74,17 @@ def validate_checkbox_field(page, values:, allowed_values:, other: false, other_
     []
   end
 
+  def validate_qualification_or_not(page, values:)
```
brucebolt: Since this only performs validation for a single form field and is specific to a question, I think it'd be better to have this method in the controller itself.  I can't see where it would be reused, and all the other methods in this helper are quite generic.

brucebolt (commented):
Great work!  I've left a few minor comments, but otherwise looks good and functionality behaves as expected when I run the branch locally.

app/controllers/coronavirus_form/offer_care_qualifications_controller.rb
Rosa-Fox: 👍

app/helpers/field_validation_helper.rb
Rosa-Fox: 👍

Rosa-Fox:
Thanks @brucebolt :) I have amended the changes.

app/controllers/coronavirus_form/offer_care_qualifications_controller.rb
```diff
@@ -10,19 +10,7 @@ def submit
...
-        allowed_values: I18n.t("coronavirus_form.questions.#{controller_name}.offer_care_type.options").map { |_, item| item.dig(:label) },
-      ) +
-      validate_checkbox_field(
-        "#{controller_name}.care_qualifications",
-        values: @form_responses[:offer_care_qualifications],
-        allowed_values: I18n.t("coronavirus_form.questions.#{controller_name}.care_qualifications.options").map { |_, item| item.dig(:label) },
-        other: @form_responses[:offer_care_qualifications_type],
-        other_field: "nursing_or_healthcare_qualification",
-      )
+    invalid_fields = validate_fields
```
brucebolt, line 13: 👍

brucebolt (approved):
Brilliant, thanks for making those changes!

PR #285, opened 2020-05-06
Disable AssetSync in Heroku

bilbof:
This will enable Heroku build to succeed since they won't try to upload assets to S3.


gclssvglx (approved):
LGTM 👍

PR #286, opened 2020-05-06
Rename hotel rooms to accomodation

leenagupte:
Trello: https://trello.com/c/f4h7AyXv

Preview App 👉 https://coronavirus-rename-hote-bs6isp.herokuapp.com/medical-equipment

# What's changed and why?

Updates the content on the "hotel rooms" and "hotel rooms number" pages to allow all kinds of accommodation and rooms to be offered.
To keep the questions consistent with the code that displays, the word "hotel" has been removed from the rest of the code base, so `hotel_rooms` has been changed to `accommodation` and `hotel_rooms_number` has been changed to `rooms_number`.

# Expected changes
## Code
- I18n lookup
- schema
- session key,
- analytics identifier
- params
- class names
- file names

## Content
Screenshot 2020-05-06 at 16 42 55
Screenshot 2020-05-06 at 16 49 43
Screenshot 2020-05-06 at 16 43 29
Screenshot 2020-05-06 at 16 43 33


config/routes.rb
```diff
@@ -32,13 +32,13 @@
...
+    # Question 2.0: Can you offer accommodation?
+    get "/accommodation", to: "accommodation#show"
+    post "/accommodation", to: "accommodation#submit"
 
-    # Question 2.1: How many hotel rooms can you offer?
-    get "/hotel-rooms-number", to: "hotel_rooms_number#show"
-    post "/hotel-rooms-number", to: "hotel_rooms_number#submit"
+    # Question 2.1: How many rooms can you offer?
+    get "/rooms-number", to: "rooms_number#show"
+    post "/rooms-number", to: "rooms_number#submit"
```
bilbof, line 41: It'd be good to think about users who are using the form when we deploy this.

If users are already on either of these pages or a previous page that links to these paths when we deploy this, the request will fail with a 404 and they'll need to refresh the page, which will then not exist, which would break the form for those users.

Deploys also take a while, so there's a chance requests made on pages served by the new code will be handled by apps running the old code.

We might want to do an A AB B deploy approach here.

Prior to shipping this, we could create a separate PR to point the new paths to the old paths? That way requests made on pages served by the new "B" code but handled by the old "A" code would continue to work. Then in this PR we could keep the old routes, but point them to the new controllers, so requests from "A" code are handled by the "B" code. Later we could remove the hotel rooms paths, once we're sure nobody is using them.

This would be a long journey, so another less time intensive approach would be to ship this at a low traffic time, and accept some users will have to restart the form.

bilbof (approved):
Looks good, just a comment to consider.

config/routes.rb
leenagupte: How much traffic does this form get? If it can be merged later in the evening, and only affect a few people, I think we should do that.

bilbof: [Very little traffic](https://grafana-paas.cloudapps.digital/d/N7NZbVrZz/coronavirus-form-services?orgId=1&from=1588767552441&to=1588770459357&var-Application=govuk-coronavirus-business-volunteer-form) compared to the vulnerable people form. Lets deploy this as-is at a low traffic time.

bilbof (approved):

bilbof:
@leenagupte this has approval from Cabinet Office. This should be deployed next week, but we'll need to let them know when we deploy it so they can make changes on their end.

PR #293, opened 2020-05-10
Use consistent quotes for translations in views

brucebolt:
In some places we are using double quotes, and in other places we are using single quotes when including translations in the views.

In the absence of view linting, updating to make this consistent in all views.


issyl0 (approved):
Eventually we'll get to consistent and correct quotes everywhere all the time. :sweat:

PR #300, opened 2020-05-13
ci: Use the `--deployment` flag on `bundle install`

issyl0:
What
----

We can simplify our CI config further by [using the config from the
Actions conventions doc](https://github.com/alphagov/govuk-developer-docs/pull/2533).

How to review
-------------

Observe that CI still works!

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,10 +1,7 @@
-name: Run tests
-
 on: [push, pull_request]
 
 jobs:
   container-jobs:
-    name: Pre-merge checks
     runs-on: ubuntu-latest
 
     services:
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -1,10 +1,7 @@
-name: Run tests
```
huwd, line 1: How come we removed the name? Not needed?

.github/workflows/ci.yml
```diff
@@ -21,36 +18,27 @@ jobs:
     steps:
     - uses: actions/checkout@v1
     - uses: ruby/setup-ruby@v1
-
-    - name: Install PostgreSQL 11 client
```
huwd, line 25: Don't we need postgres to run the tests?

.github/workflows/ci.yml
```diff
@@ -21,36 +18,27 @@ jobs:
...
       with:
         path: vendor/bundle
-        key: ${{ runner.os }}-gems-${{ hashFiles('**/Gemfile.lock') }}
-        restore-keys: |
-          ${{ runner.os }}-gems-
+        key: bundle-${{ hashFiles('**/Gemfile.lock') }}
+        restore-keys: bundle
 
-    - name: Build App
+    - run: sudo apt-get -yqq install libpq-dev
```
huwd, line 27: Interesting, why move this above the env?

huwd (approved):
Left a couple of questions for bits i didn't understand but looks good

.github/workflows/ci.yml
issyl0: To keep the `uses` together.

.github/workflows/ci.yml
issyl0: Yeah. We've broadly decided on no name in the conventions doc (see the link in the PR description).

PR #307, opened 2020-05-14
Use exclusive checkboxes

alex-ju:
### What

Use 'exclusive' checkboxes on 'expert-advice-type' and 'offer-care-qualifications' pages, by using a single fieldset for all items and de-select items when the 'no/none' option is selected.

~~This PR is still draft as it's dependent on us [being able to import component scripts individually](https://github.com/alphagov/govuk_publishing_components/pull/1472) hence **not yet ready for review** and with a test failing.~~

### How to review

Check those pages in your browser/device of choice or check the code and trust the screen captures below.

<table>
<tr><th>Before</th><th>After</th></tr>
<tr><td valign="top">

localhost_5000_expert-advice-type (1)

</td><td valign="top">

localhost_5000_expert-advice-type

</td></tr>
<tr><td valign="top">

localhost_5000_offer-care-qualifications

</td><td valign="top">

localhost_5000_offer-care-qualifications (1)

</td></tr>
<tr><td valign="top">

n/a

</td><td valign="top">
JavaScript disabled

localhost_5000_expert-advice-type

</td></tr>
<tr><td valign="top">

n/a

</td><td valign="top">

JavaScript disabled
localhost_5000_offer-care-qualifications

</td></tr>
</table>

### Links

[Trello](https://trello.com/c/kJe25doD) [cards](https://trello.com/c/TUIIAq8f)

### DIFF
--- a/app/controllers/coronavirus_form/expert_advice_type_controller.rb
+++ b/app/controllers/coronavirus_form/expert_advice_type_controller.rb
@@ -15,6 +15,8 @@ class CoronavirusForm::ExpertAdviceTypeController < ApplicationController
         values: @form_responses[:expert_advice_type],
         allowed_values: I18n.t("coronavirus_form.questions.#{controller_name}.options").map { |_, item| item.dig(:label) },
         other: @form_responses[:expert_advice_type_other],
+        exclusive: selected_exclusive?,
+        exclusive_values: exclusive_values,
       )
 
     if invalid_fields.any?
@@ -50,6 +52,14 @@ private
     @form_responses[:expert_advice_type].include?(
       I18n.t("coronavirus_form.questions.#{controller_name}.options.other.label"),
     )
+  end
+
+  def selected_exclusive?
+    @form_responses[:expert_advice_type].include?(exclusive_values.first)
+  end
+
+  def exclusive_values
+    I18n.t("coronavirus_form.questions.#{controller_name}.options").map { |_, item| item.dig(:label) if item.dig(:exclusive) == true }.compact
   end
 
   def previous_path
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -1,11 +1,20 @@
 //= require jquery
-// = require govuk/all.js
-//= require govuk_publishing_components/components/cookie-banner
+//= require govuk_publishing_components/modules
 //= require govuk_publishing_components/lib/cookie-functions
+//= require govuk_publishing_components/components/button
+//= require govuk_publishing_components/components/character-count
+//= require govuk_publishing_components/components/checkboxes
+//= require govuk_publishing_components/components/cookie-banner
+//= require govuk_publishing_components/components/error-summary
+//= require govuk_publishing_components/components/radio
 //= require analytics
 //= require cookies
 window.CookieSettings.start()
-window.GOVUKFrontend.initAll()
+
 if (window.GOVUK.analyticsInit) {
   window.GOVUK.analyticsInit()
 }
+
+$(document).ready(function () {
+  window.GOVUK.modules.start()
+})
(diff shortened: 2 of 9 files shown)


app/controllers/coronavirus_form/expert_advice_type_controller.rb
```diff
@@ -15,6 +15,8 @@ def submit
         values: @form_responses[:expert_advice_type],
         allowed_values: I18n.t("coronavirus_form.questions.#{controller_name}.options").map { |_, item| item.dig(:label) },
         other: @form_responses[:expert_advice_type_other],
+        exclusive: selected_exclusive?,
+        exclusive_values: I18n.t("coronavirus_form.questions.#{controller_name}.options").map { |_, item| item.dig(:label) if item.dig(:exclusive) == true },
```
leenagupte: It looks like these `exclusive_values` are the same as the value used to determine `selected_exclusive?`.

Can this be DRY-ed up, so it obvious we're looking at the same thing? Perhaps move finding the `exclusive_values` into method, and have `selected_exclusive?` call `exclusive_values.first` ?

Also, as it looks like these methods and values are needed in multiple controllers, they could be moved into a helper.

leenagupte (commented):
This looks good. I only have one suggestion for DRY-ing up the code a bit, otherwise 👍

app/controllers/coronavirus_form/expert_advice_type_controller.rb
alex-ju: Good suggestion, @leenagupte. I updated the code now and rebased again to master. We only use them in 2 places atm, I'm happy to create a helper if we'll need them elsewhere.

leenagupte:
@alex-ju One other thing about child text boxes.  They don't close when the user selects the exclusive option, and the data typed into them is still being saved.

Screenshot 2020-05-21 at 17 45 01

alex-ju:
> One other thing about child text boxes. They don't close when the user selects the exclusive option, and the data typed into them is still being saved.

Indeed they don't, and this is because in [the conditionally revealed content implementation in `govuk-frontend`](https://github.com/alphagov/govuk-frontend/blob/master/src/govuk/components/checkboxes/checkboxes.js#L36) we only listen for `click` events and not property changes too. I intend to raise this with the Design System team, but until (and if) this gets in and released it will take a while.

Regarding the content in textarea, I would be reluctant on removing that content via JavaScript for two reasons:
- if you accidentally click "I do not have a qualification" you'll lose the textarea content
- the 'exclusive' script will never be generic enough to figure out whatever content is placed within the conditionally revealed container (it could be a mix of inputs with different states).

What do you say if we clean it on the backend, after form submission, if "I do not have a qualification" is selected?

leenagupte:
> What do you say if we clean it on the backend, after form submission, if "I do not have a qualification" is selected?

@alex-ju This sounds good. I think it's important that we don't accidentally save any "extra" information.

alex-ju:
> > What do you say if we clean it on the backend, after form submission, if "I do not have a qualification" is selected?
> 
> @alex-ju This sounds good. I think it's important that we don't accidentally save any "extra" information.

@leenagupte I checked now and this scenario is already captured in [`update_session_store`](https://github.com/alphagov/govuk-coronavirus-business-volunteer-form/blob/master/app/controllers/coronavirus_form/expert_advice_type_controller.rb#L38) for this controller. I rebased it again against master.

leenagupte:
> update_session_store

@alex-ju That looks like it's only for the construction and it services questions in the expert_advice controller, not the qualifications. You'll need something similar in the [offer_care_qualifications controller](app/controllers/coronavirus_form/offer_care_qualifications_controller.rb)

alex-ju:
> @alex-ju That looks like it's only for the construction and it services questions in the expert_advice controller, not the qualifications. You'll need something similar in the [offer_care_qualifications controller](app/controllers/coronavirus_form/offer_care_qualifications_controller.rb)

Right, I was under the impression all conditional reveals are checked in a similar fashion. Added a similar logic to the `offer_care_qualifications_controller`.

leenagupte (approved):
💯 🎉

PR #322, opened 2020-05-20
Task/autocreate schema from meta schema

danwilliams:
## Summary of changes

The approach described here is different from what the ticket suggested under the **Considerations** section but it does meet the **Acceptance** **criteria** of the ticket. i.e one can (under this scheme) now change the content only once and regenerating the schema once.

In addition this approach avoids having to refactor the en.yml file, and the rest of the codebase de-risking this change. It also minimises the structural constraints placed on the locale file, and allows the schema to be specified in a file with a near identical syntax to that of a json-schema file, which allows for linting and validation using that toolchain.

Initially the approach taken did follow the approach outlined in the **Considerations** section 
of the ticket. That experiment reached a point where the structure of the questions in the yml file had been flattened. 

I've adopted a scheme whereby the form response schema is automatically generated from the meta-schema `config/form_response_metaschema.json` and the `en.yml` locale file whenever rake is invoked. i.e when the server is started, or the tests are ran. (Note that the following is replicated from additions to the README file in the branch)

The generated file appears at:
```
config/schemas/form_response.json
```

If you want to generate it manually you can run:
```
bundle exec rake generate-schema
```

If you want to edit the schema, edit the meta-schema file:
```
config/form_response_metaschema.json
```

This json file is almost identical to a json schema, the only difference is
that it goes through a processing step that replaces any enums with the prefix
`meta_i18n_enum:` with values found in `config/locales/en.yml`.

### Example

Specifying an enum like so:

```
"enum" [
    "meta_i18n_enum:department.options"
]
```

Will result in the processing step looking at `department.options` in `en.yml`
path, and adding each of the labels of the options to the enum. i.e When the
above enum is processed together with an `en.yml` file with the following
structure:

```
department:
    options:
      medical:
        label: "Medical"
      engineering:
        label: "Engineering"
```

The enum in the schema will look like the following:

```
"enum" [
    "Medical",
    "Engineering",
]
```

## How to review

One can run:

```
bundle exec rake generate-schema
```

(or any rake command)

and observe that the form response schema is generated at `/config/schemas/form_response.json`

Diffing that file against the one in master shows that nothing substantial has changed in the file (a few formatting changes, and a (now correctly) escaped backslash.

Similarly one can run the tests and see that the file is correctly loaded, and that the validation is ran.

## Links

https://trello.com/c/7Ymm07ol/387-spike-generating-the-json-schema-using-erb

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,6 +23,7 @@
 
 # Ignore master key for decrypting credentials and more.
 /config/master.key
+/config/schemas/form_response.json
 
 coverage
 tmp
--- a/Rakefile
+++ b/Rakefile
@@ -7,5 +7,6 @@
 require_relative "config/application"
 
 Rails.application.load_tasks
+Rake::Task["generate-schema"].invoke
 
 task default: %i[spec lint]
--- a/lib/tasks/generate_schema.rake
+++ b/lib/tasks/generate_schema.rake
@@ -1,0 +1,38 @@
+require "fileutils"
+
+def replace_meta_enums json
+  json.each do |key, value|
+    if (key == "enum") && value.any? { |member| member.start_with?("meta_i18n_enum:") }
+      if value.size > 1
+        throw ArgumentError("A meta enum should have a length of one. Found: #{value}")
+      end
+      json["enum"] = I18n.t(value.first.split(":", 2).last).map { |_, option| option[:label] }
+    elsif value.class == Hash
+      replace_meta_enums value
+    elsif value.class == Array
+      value.each do |member|
+        handle_array member
+      end
+    end
+  end
+end
+
+def handle_array member
+  if member.class == FalseClass
+    replace_meta_enums member
+  elsif member.class == Array
+    member.each do |nested_member|
+      handle_array nested_member
+    end
+  end
+end
+
+desc "Generate json schema from locale and meta schema files"
+task "generate-schema" => :environment do
+  schema = replace_meta_enums JSON.parse(File.read(Rails.root.join("config/form_response_meta_schema.json")))
+  schema.delete("description")
+  FileUtils.mkdir_p "config/schemas"
+  File.open("config/schemas/form_response.json", "w") do |f|
+    f.write(JSON.pretty_generate(schema))
+  end
+end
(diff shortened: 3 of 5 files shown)


issyl0:
Needs a rebase, please.

.gitignore
```diff
@@ -23,6 +23,7 @@
 
 # Ignore master key for decrypting credentials and more.
 /config/master.key
+/config/schemas/form_response.json
```
bilbof, line 26: I'd prefer that the schema be committed, rather than generating it on the fly.

This allows us to see in PRs whether the schema has changed, and if we need to inform stakeholders of the fact.
I hope that would also draw developers attention to schema changes, so they could plan deploys accordingly.

Rakefile
```diff
@@ -7,5 +7,6 @@
 require_relative "config/application"
 
 Rails.application.load_tasks
+Rake::Task["generate-schema"].invoke
```
bilbof, line 10: If possible, could we remove this line?

I'd prefer the schema to be regenerated manually by a developer, rather than as a side effect of other rake tasks. Regenerating the schema in this way would probably be confusing, and will be unnecessary for many rake tasks.

lib/tasks/generate_schema.rake
```diff
@@ -0,0 +1,46 @@
...
+  elsif member.class == Array
+    member.each do |nested_member|
+      handle_array nested_member
+    end
+  end
+end
+
+desc "Generate json schema from locale and meta schema files"
+task "generate-schema" => :environment do
+  schema = replace_meta_enums JSON.parse(File.read(Rails.root.join("config/form_response_meta_schema.json")))
```
bilbof, line 40: It would be great if we could figure out a way to generate the entire schema (all keys and values) from the yaml file, rather than just those enums.

I can see situations where we'll add a key or value to the yaml file but forget to add it to the schema. Then, when we enforce the schema, that would prevent people from submitting the form. Ideally this work will prevent that from happening, but perhaps that'd require refactoring the yaml file so we read more from it.

bilbof (changes requested):
Good start 👍 

I've left a couple of comments to be addressed before we merge this.

leenagupte:
Will be used as a reference for on-going work.

PR #327, opened 2020-05-20
S3 error pages

bilbof:
This will upload error pages to an S3 bucket defined in https://github.com/alphagov/covid-engineering/pull/229.

This will occur, similar to assets (with AssetSync) during initialization of the apps.

It will mean we do several writes to AWS S3 during the startup, which will slow down deploys slightly.

Co-Authored-By: @GDSNewt
Co-Authored-By: @steveholmes

Still to do:

1. After this is deployed to prod, enable some CloudFront rules in production 
1. Do the same for the other apps

https://trello.com/c/di3xSyYN/385-serve-error-pages-from-s3


bilbof:
Looks like we need to fix Heroku, too: it's running `error_pages:upload` when it shouldn't be.

issyl0:
I'll take a look at this today.

spec/lib/upload_error_pages_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+        lambda { |context|
+          bucket[context.params[:key]] = { body: context.params[:body] }
+          {}
+        },
+      )
+
+      responses = UploadErrorPages.new(
+        error_pages: %w[404.html 500.html],
+        s3_client: client,
+        bucket_name: "my-lovely-bucket-name",
```
Rosa-Fox, line 24: great name :)

Rosa-Fox:
It looks ok to me. Might be good for @issyl0 to give a look over once they return. Otherwise am happy to approve!

issyl0 (approved):

PR #340, opened 2020-05-26
Alert Slack when Concourse tasks fail

brucebolt:
What
----

Posts a message to #govuk-corona-services-tech on Slack when Concourse tasks fail.

Makes the same change as in https://github.com/alphagov/govuk-coronavirus-vulnerable-people-form/pull/364.

Example:
Screenshot 2020-05-26 at 14 00 11

Links
-----

Trello card: https://trello.com/c/TSaLQkQp


bilbof (approved):
🏅

PR #342, opened 2020-05-27
Remove unrequired fields

bilbof:
What
----

Similar to https://github.com/alphagov/govuk-coronavirus-vulnerable-people-form/pull/356, this removes some unnecessary fields from new `FormResponse` records.

It also changes the structure of the data we save in `form_response` from an array to a hash/json. Since we control how we export the data from the DB, we are able to do this without changing the data export format for our data team downstream.

Links
-----

https://trello.com/c/zzS6yMZG/493-remove-unnecessary-fields-from-formresponse

### DIFF
--- a/spec/controllers/coronavirus_form/check_answers_controller_spec.rb
+++ b/spec/controllers/coronavirus_form/check_answers_controller_spec.rb
@@ -4,6 +4,7 @@ require "spec_helper"
 
 RSpec.describe CoronavirusForm::CheckAnswersController, type: :controller do
   include FormResponseHelper
+  include SchemaHelper
   include_examples "session expiry"
 
   let(:current_template) { "coronavirus_form/check_answers" }
@@ -27,12 +28,39 @@ RSpec.describe CoronavirusForm::CheckAnswersController, type: :controller do
     end
 
     it "saves the form response to the database" do
-      session["attribute"] = "key"
+      data = valid_data
+      session.merge!(data)
       post :submit
 
-      expect(FormResponse.first.form_response).to eq(
-        [%w[attribute key], %w[reference_number abc]],
-      )
+      expected_data = data.deep_stringify_keys.merge("reference_number" => "abc")
+      expect(FormResponse.first.form_response).to eq(expected_data)
+    end
+
+    context "when unwanted fields are present in the session" do
+      it "excludes those fields from the FormResponse" do
+        data = valid_data
+        session.merge!(data)
+        session["_csrf_token"] = "1234"
+        session["current_path"] = "/current"
+        session["previous_path"] = "/previous"
+        session["check_answers_seen"] = true
+        session["session_id"] = "12345"
+        post :submit
+
+        expected_data = data.deep_stringify_keys.merge("reference_number" => "abc")
+        expect(FormResponse.first.form_response).to eq(expected_data)
+        expect(validate_against_form_response_schema(FormResponse.first.form_response)).to be_empty
+      end
+    end
+
+    it "provides the data in an expected exportable format" do
+      data = valid_data
+      session.merge!(data)
+      post :submit
+
+      expect(validate_against_form_response_schema(FormResponse.last.form_response)).to be_empty
+
+      expect(JSON.parse(FormResponse.last.form_response.to_json).to_h).to eq(data.merge("reference_number" => "abc").deep_stringify_keys)
     end
 
     it "sends an email" do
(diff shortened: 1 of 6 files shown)


spec/controllers/coronavirus_form/check_answers_controller_spec.rb
```diff
@@ -28,13 +29,45 @@
...
+        "contact_details" => { "contact_name" => "Harry Potter" },
+        "medical_equipment" => "Yes",
+        "reference_number" => "abc",
       )
     end
 
+    context "when unwanted fields are present in the session" do
+      it "excludes those fields from the FormResponse" do
+        data = valid_data
+        data.map { |k, v| session[k.to_s] = v }
```
leenagupte: What is this line doing?

leenagupte: Oh! Are you setting the session values?

bilbof: This is adding the valid_data from the schema helper to the session, with stringified keys.

leenagupte: Why does it need `to_s`?  Does `session.merge!(valid_data)` work? Or  `session.merge!(valid_data.deep_stringify_keys)`

bilbof: Yeah that works. That's way better, thanks!

spec/controllers/coronavirus_form/check_answers_controller_spec.rb
```diff
@@ -28,13 +29,45 @@
...
+      session["medical_equipment"] = I18n.t("coronavirus_form.questions.medical_equipment.options.option_yes.label")
+      session["contact_details"] = { "contact_name" => "Harry Potter" }
       post :submit
 
       expect(FormResponse.first.form_response).to eq(
-        [%w[attribute key], %w[reference_number abc]],
+        "attribute" => "key",
+        "contact_details" => { "contact_name" => "Harry Potter" },
+        "medical_equipment" => "Yes",
+        "reference_number" => "abc",
```
leenagupte: I wonder if this test could use `valid_data` as well? So as below, use `session.merge!(valid_data)`  in the set-up, in in the expectation use:

`expect(FormResponse.first.form_response).to eq(valid_data.merge(reference_number: "abc").deep_stringify_keys)`

spec/controllers/coronavirus_form/check_answers_controller_spec.rb
```diff
@@ -28,13 +29,45 @@
...
       )
     end
 
+    context "when unwanted fields are present in the session" do
+      it "excludes those fields from the FormResponse" do
+        data = valid_data
```
leenagupte, line 41: This assignment isn't really need. `valid_data` could be used instead.

spec/controllers/coronavirus_form/check_answers_controller_spec.rb
bilbof: This test is probably not necessary, since it's a duplicate of the other test - I'll remove it

spec/helpers/schema_helper_spec.rb
```diff
@@ -585,29 +585,29 @@
...
+      it "does not allow check_answers_seen to be stored" do
         data = valid_data.merge(check_answers_seen: true)
-        expect(validate_against_form_response_schema(data)).to be_empty
+        expect(validate_against_form_response_schema(data).first).to include("check_answers_seen")
       end
 
       it "returns a list of errors when check_answers_seen has an invalid value" do
         data = valid_data.merge(check_answers_seen: "Foo")
-        expect(validate_against_form_response_schema(data).first).to include("check_answers_seen")
+        expect(validate_against_form_response_schema(data).first).to include("[\"check_answers_seen\"]")
```
leenagupte: Is this test still needed?

spec/controllers/coronavirus_form/check_answers_controller_spec.rb
bilbof: `valid_data` is not deterministic, since it uses `SecureRandom.uuid` and `[].sample`, so I use the assignment so that I can use the same `data` in the expectation and the form submission.

leenagupte: I forgot about the SecureRandoms. I wonder who put them there? 🤔 😊

leenagupte (commented):
Looks good! I just have a couple of queries about the tests, but nothing major.

spec/helpers/schema_helper_spec.rb
bilbof: No, I'll remove it, thanks!

leenagupte (approved):
🏅 LGTM!

PR #344, opened 2020-05-28
Make whole_number fields an integer

steveholmes:
What
----

The `whole_number` part of the schema has been updated to be an integer
Form processing has been changed to parse whole number inputs to be integers.

Why
---

We are passing numbers as strings to the Cabinet Office data team.
They would like us to pass integers rather than strings.

How to review
-------------

Fields affected are:-
Product Details page:
`Lead Time` and `Product Quantity`

Offer Staff page:
All staff numbers

These fields should be tested for validation and the FormResponse in the DB has integer values for these fields.

Links
-----

https://trello.com/c/YW5cmUu9/485-make-wholenumber-fields-an-integer


issyl0 (approved):
LGTM. But we don't have approval from Josh and Tristan and I'm loathe to email them on a Sunday. Even though this is a thing they wanted, we should probably follow protocol. One for tomorrow!

app/helpers/product_helper.rb
```diff
@@ -23,4 +23,18 @@ def products_except(product_id)
...
+    product[:lead_time] = safe_integer_cast(product[:lead_time])
+    product
+  end
+
+private
+
+  def safe_integer_cast(number)
+    Integer(number)
+  rescue ArgumentError, TypeError
+    nil
```
bilbof, line 38: I wonder if we want to raise an error here, and get the user to give us something we can coerce into an integer?

Otherwise, we might end up putting records into the DB without product lead time or quantity provided.

spec/controllers/coronavirus_form/offer_staff_type_controller_spec.rb
```diff
@@ -58,14 +58,14 @@
...
-        translators_number: params[:translators_number],
-        other_staff_number: params[:other_staff_number],
+        cleaners_number: Integer(params[:cleaners_number]),
+        developers_number: Integer(params[:developers_number]),
+        medical_staff_number: Integer(params[:medical_staff_number]),
+        office_staff_number: Integer(params[:office_staff_number]),
+        security_staff_number: Integer(params[:security_staff_number]),
+        trainers_number: Integer(params[:trainers_number]),
+        translators_number: Integer(params[:translators_number]),
+        other_staff_number: Integer(params[:other_staff_number]),
```
bilbof, line 68: I wonder if any of these fields are permitted to be `nil`, which I suppose would raise an unhandled `TypeError` here and a 500 to the user.

app/helpers/product_helper.rb
steveholmes: We do validate earlier and re-request from the user.
Then we cast the responses once all validation passes. So, it **should** be impossible to get an error at this point.

spec/controllers/coronavirus_form/offer_staff_type_controller_spec.rb
steveholmes: For handling these values in the form I actually use the the `String#to_i` which defaults to 0 rather than error/nil, so they should never have a 500 here.

PR #352, opened 2020-06-02
Bugfix: additional product back link

bilbof:
This fixes an issue with the back link on the additional product page: this was changed from a String to a Symbol which introduced a bug.

I've added a test to catch this for next time.

https://trello.com/c/6fdlmZlR/509-prefill-previous-answers-on-medical-equipment-page

### DIFF
--- a/app/controllers/coronavirus_form/additional_product_controller.rb
+++ b/app/controllers/coronavirus_form/additional_product_controller.rb
@@ -35,7 +35,7 @@ private
 
   def previous_path
     session["product_details"] ||= []
-    latest_product_id = (session["product_details"].last || {}).dig("product_id")
+    latest_product_id = session["product_details"].last&.with_indifferent_access&.dig("product_id")
     product_details_url(product_id: latest_product_id)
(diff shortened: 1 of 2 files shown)


app/controllers/coronavirus_form/additional_product_controller.rb
```diff
@@ -35,7 +35,7 @@ def update_session_store
 
   def previous_path
     session["product_details"] ||= []
-    latest_product_id = (session["product_details"].last || {}).dig("product_id")
+    latest_product_id = (session["product_details"].last || {}).with_indifferent_access.dig("product_id")
```
leenagupte: I know you didn't add it, but rather than using a "OR", the safe navigation operator could be used:

```
latest_product_id = session["product_details"]&.with_indifferent_access.dig("product_id")
```

bilbof: Thanks!

leenagupte (approved):
💯

PR #361, opened 2020-06-04
Redact IDs from urls for cross domain tracking

Rosa-Fox:
What
----

Redact the `product_id` and `reference_number` params for Google Analytics X domain tracking.

Technically this isn't personally identifiable information (PII) on its own, but it could be along with the backend data.

How to review
-------------

:ship: Since this service is continuously deployed, all testing must be done
before the pull request is merged. :ship:

Describe the steps required to review and test the changes.

- I'm not hugely familiar with JS so any tips on refactoring the logic could be good. I am thinking it may be good to move the functions out into a separate file?

for the regexs:
- the `product_id` is created using `SecureRandom.uuid`.
- the `reference_number` is created using a timestamp and random number [see here](https://github.com/alphagov/govuk-coronavirus-business-volunteer-form/blob/master/app/controllers/coronavirus_form/check_answers_controller.rb#L64).
- The Google doc linked to the Trello card may also be useful for giving some more context.

Links
-----

[Trello](https://trello.com/c/JFMD3vzV/501-dont-send-google-analytics-user-transaction-id-query-parameters)

### DIFF
--- a/app/assets/javascripts/analytics.js.erb
+++ b/app/assets/javascripts/analytics.js.erb
@@ -1,6 +1,20 @@
 // Start cookie banner (display settings controlled internally by cookies)
 var element = document.querySelector('[data-module="cookie-banner"]')
+var PRODUCT_ID_PATTERN = /\b[0-9a-f]{8}\b-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-\b[0-9a-f]{12}\b/
+var REFERENCE_NUMBER_PATTERN = /\d{8}-\d{6}-.{6}/
+
 new GOVUK.Modules.CookieBanner().start($(element))
+
+var stripPII = function(url) {
+  url = url.replace(REFERENCE_NUMBER_PATTERN, 'reference_number')
+  url = url.replace(PRODUCT_ID_PATTERN, 'product_id')
+
+  if(url.includes('reference_number' || 'product_id')) {
+    return url
+  } else {
+    return location.protocol + '//' + location.host + location.pathname
+  }
+}
 
 var analyticsInit = function() {
   <% if Rails.application.config.analytics_tracking_id.present? %>
@@ -19,6 +33,7 @@ var analyticsInit = function() {
       ga('create', trackingId, 'auto')
       ga('set', 'allowAdFeatures', false);
       ga('set', 'anonymizeIp', true);
+      ga('set', 'location', stripPII(window.location.href));
       ga('send', 'pageview');
 
       // Cross Government Domain Google Analytics
@@ -27,6 +42,7 @@ var analyticsInit = function() {
       ga('govuk_shared.set', 'anonymizeIp', true);
       ga('govuk_shared.set', 'allowAdFeatures', false);
       ga('govuk_shared.linker:autoLink', ['www.gov.uk']);
+      ga('govuk_shared.set', 'location', stripPII(window.location.href));
       ga('govuk_shared.send', 'pageview');
     }
   <% end %>


app/assets/javascripts/analytics.js.erb
```diff
@@ -27,6 +49,7 @@ var analyticsInit = function() {
       ga('govuk_shared.set', 'anonymizeIp', true);
       ga('govuk_shared.set', 'allowAdFeatures', false);
       ga('govuk_shared.linker:autoLink', ['www.gov.uk']);
+      ga('govuk_shared.set', 'location', stripPII(window.location.href));
```
bilbof, line 45: Is there a way to remove all query parameters from the URLs we send GA? Just in case these get renamed in future or we add others that are also PII but forget to strip them here.

Rosa-Fox: @bilbof I did think this, but the technical guidance directed to replace the params with strings so I followed that. I will check with a performance analyst to see if there is any reason for preserving the query strings, otherwise I can remove them.

Rosa-Fox: @bilbof from Harriet:

"So it's actually better to have [redacted-number] because it shows that there is missing data and what the redacted data is". I'll look into ways of testing it now.

bilbof: Thanks @Rosa-Fox!

Rosa-Fox: @bilbof I looked into testing with Jest... it seems possible but we will need to update our pipeline to run the JS tests. I will write a separate card so that we can get this part merged quicker.

bilbof:
@Rosa-Fox are you ready for re-review on this one? 😃

app/assets/javascripts/analytics.js.erb
```diff
@@ -2,6 +2,28 @@
...
 new GOVUK.Modules.CookieBanner().start($(element))
 
+var referenceNumberPattern = function() {
+  var pattern = /\d{8}-\d{6}-.{6}/
+  return pattern
+}
+
+var productIdPattern = function() {
+  var pattern = /\b[0-9a-f]{8}\b-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-\b[0-9a-f]{12}\b/
+  return pattern
```
bilbof: You could use `var PRODUCT_PATTERN = /\b[0-9a-f]{8}\...` at the top of the file, since this can be treated as a constant. Likewise for the `REF_NUMBER_PATTERN`.

app/assets/javascripts/analytics.js.erb
```diff
@@ -2,6 +2,28 @@
...
+}
+
+var stripPII = function(url) {
+  if (referenceNumberPattern().test(url)) {
+    var stripped = url.replace(referenceNumberPattern(), 'reference_number')
```
bilbof: Rather than setting stripped here, just return early e.g. `return url.replace(REF_NUMBER_PATTERN, 'reference_number')`. Then the `else` statement later becomes unnecessary and you can do `return url` at the end.

bilbof (approved):
Couple of refactorings requested but feel free to merge 👏

app/assets/javascripts/analytics.js.erb
bilbof: This is how you might get the URL without query params: https://stackoverflow.com/a/6257480

app/assets/javascripts/analytics.js.erb
Rosa-Fox: 👍

app/assets/javascripts/analytics.js.erb
Rosa-Fox: 👍 way better!

app/assets/javascripts/analytics.js.erb
```diff
@@ -1,7 +1,17 @@
...
+var PRODUCT_ID_PATTERN = /\b[0-9a-f]{8}\b-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-\b[0-9a-f]{12}\b/
+var REFERENCE_NUMBER_PATTERN = /\d{8}-\d{6}-.{6}/
+
 new GOVUK.Modules.CookieBanner().start($(element))
 
+var stripPII = function(url) {
+  url = url.replace(REFERENCE_NUMBER_PATTERN, 'reference_number')
+  url = url.replace(PRODUCT_ID_PATTERN, 'product_id')
+
+  return url
```
bilbof: I'd recommend changing this function to return `location.protocol + '//' + location.host + location.pathname`, unless we want other query params to be sent to GA.

bilbof (approved):

app/assets/javascripts/analytics.js.erb
Rosa-Fox: @bilbof I have amended.

bilbof (approved):
Looks good! 🎉

PR #362, opened 2020-06-05
Back links

JonathanHallam:
This fixes all the back links for this service. A number of them weren't working properly if a user had said that they could offer support. 

Trello - https://trello.com/c/Efx2SrSn/510-back-button-should-take-users-back-to-the-page-they-were-on-directly-before


GDSNewt (approved):
LGTM

PR #381, opened 2020-06-15
Make company number mandatory

gclssvglx:
Why
----

CCS can only process offers from UK companies who have a company number, so we don't want UK companies who don't have a UK Company Registration Number (CRN) completing the form.

Non-UK companies can still complete the form without a company number.

What
----

This change replaces company number (optional) with a radio option that asks the user if their company is registered in the UK. If 'yes' is selected, it asks the user to enter their company number as a mandatory field. If 'no' is selected, it does not ask the user for their company
number.

Show suitable error messages and update the check your answers screen to include the new field and company number (if appropriate).

Links
-----

[Trello card](https://trello.com/c/JT4SfRoQ/576-make-company-number-mandatory)

Screenshots
-----

**New business details page**

business-details-001-implemented

**New business details page with nothing entered**

screencapture-localhost-5000-business-details-2020-06-16-08_46_24

**New business details page with UK registered and UK location selected**

screencapture-localhost-5000-business-details-2020-06-16-08_47_10

**New check your details page showing UK registered and company number**

check-your-answers-001-implemented

**New check your details page showing not-UK registered and no company number**

check-your-answers-002-implemented

### DIFF
--- a/app/controllers/coronavirus_form/business_details_controller.rb
+++ b/app/controllers/coronavirus_form/business_details_controller.rb
@@ -1,13 +1,15 @@
 # frozen_string_literal: true
 
 class CoronavirusForm::BusinessDetailsController < ApplicationController
-  REQUIRED_FIELDS = %w[company_name].freeze
+  REQUIRED_FIELDS = %w[company_name company_is_uk_registered].freeze
   TEXT_FIELDS = %w[company_name company_number].freeze
 
   def submit
     @form_responses = {
       business_details: sanitized_business_details(params),
     }
+
+    @form_responses[:business_details][:company_number] = "" if @form_responses[:business_details][:company_is_uk_registered] == I18n.t("coronavirus_form.questions.business_details.company_is_uk_registered.options.not_united_kingdom.label")
 
     invalid_fields = validate_fields
 
@@ -46,12 +48,21 @@ private
                 validate_postcode("company_postcode", @form_responses[:business_details][:company_postcode])
               end
 
+    errors << if @form_responses[:business_details][:company_is_uk_registered] == I18n.t("coronavirus_form.questions.business_details.company_is_uk_registered.options.united_kingdom.label") &&
+        @form_responses[:business_details][:company_number].blank?
+                { field: "company_number",
+                  text: t("coronavirus_form.questions.#{controller_name}.company_is_uk_registered.options.united_kingdom.company_number.custom_error") }
+              elsif @form_responses[:business_details][:company_is_uk_registered] == I18n.t("coronavirus_form.questions.business_details.company_is_uk_registered.options.united_kingdom.label")
+                validate_company_number(controller_name, @form_responses[:business_details][:company_number])
+              end
+
     errors.compact.flatten
   end
 
   def sanitized_business_details(params)
     {
       company_name: strip_tags(params[:company_name]).presence,
+      company_is_uk_registered: strip_tags(params[:company_is_uk_registered]).presence,
       company_number: strip_tags(params[:company_number]).presence,
       company_size: strip_tags(params[:company_size]).presence,
       company_location: strip_tags(params[:company_location]).presence,
--- a/app/helpers/check_answers_helper.rb
+++ b/app/helpers/check_answers_helper.rb
@@ -185,7 +185,8 @@ module CheckAnswersHelper
       joiner = "<br>"
     elsif question.eql?("business_details")
       concatenated_answer << "Company name: #{answer[:company_name]}" if answer[:company_name]
-      concatenated_answer << "Company number: #{answer[:company_number]}" if answer[:company_number]
+      concatenated_answer << "Company registered in the UK: #{answer[:company_is_uk_registered]}"
+      concatenated_answer << "Company number: #{answer[:company_number]}" if answer[:company_number].present?
       concatenated_answer << "Company size number: #{answer[:company_size]}" if answer[:company_size]
       concatenated_answer << "Company location: #{answer[:company_location]}" if answer[:company_location]
       concatenated_answer << "Company postcode: #{answer[:company_postcode]}" if answer[:company_postcode]
--- a/app/helpers/field_validation_helper.rb
+++ b/app/helpers/field_validation_helper.rb
@@ -102,6 +102,14 @@ module FieldValidationHelper
     end
   end
 
+  def validate_company_number(field, company_number)
+    if company_number =~ /^([0-9]{8}|[A-Za-z]{2}[0-9]{6})$/
+      []
+    else
+      [{ field: field.to_s.sub(".", "_"), text: t("coronavirus_form.errors.company_number") }]
+    end
+  end
+
   def validate_charge_field(field, value)
     if value.blank?
       return [{
--- a/app/views/coronavirus_form/business_details.html.erb
+++ b/app/views/coronavirus_form/business_details.html.erb
@@ -22,6 +22,7 @@
 <%= render "govuk_publishing_components/components/input", {
   id: "company_name",
   name: "company_name",
+  heading_size: "s",
   label: {
     text: t("coronavirus_form.questions.business_details.company_name.label"),
   },
--- a/app/views/coronavirus_form/business_details.html.erb
+++ b/app/views/coronavirus_form/business_details.html.erb
@@ -29,13 +29,35 @@
   value: @form_responses.dig(:business_details, :company_name),
 } %>
 
-<%= render "govuk_publishing_components/components/input", {
-  name: "company_number",
-  label: {
-    text: t("coronavirus_form.questions.business_details.company_number.label"),
-  },
-  error_message: error_items('company_number'),
-  value: @form_responses.dig(:business_details, :company_number),
+<%= render "govuk_publishing_components/components/radio", {
+  heading: t("coronavirus_form.questions.business_details.company_is_uk_registered.label"),
+  heading_size: "s",
+  name: "company_is_uk_registered",
+  error_message: error_items('company_is_uk_registered'),
+  items: [
+    {
+      id: "company_is_uk_registered",
+      value: t("coronavirus_form.questions.business_details.company_is_uk_registered.options.united_kingdom.label"),
+      text: t("coronavirus_form.questions.business_details.company_is_uk_registered.options.united_kingdom.label"),
+      checked: @form_responses.dig(:business_details, :company_is_uk_registered) == t("coronavirus_form.questions.business_details.company_is_uk_registered.options.united_kingdom.label"),
+      conditional: render("govuk_publishing_components/components/input", {
+        label: {
+          text: t("coronavirus_form.questions.business_details.company_is_uk_registered.options.united_kingdom.company_number.label"),
+        },
+        id: "company_number",
+        name: "company_number",
+        hint: t("coronavirus_form.questions.business_details.company_is_uk_registered.options.united_kingdom.company_number.hint"),
+        error_message: error_items('company_number'),
+        value: @form_responses.dig(:business_details, :company_number),
+        width: 10
+      })
+    },
+    {
+      value: t("coronavirus_form.questions.business_details.company_is_uk_registered.options.not_united_kingdom.label"),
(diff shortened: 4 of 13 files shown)


simonwhatley:
On the business details form, the error message for the "Is your company registered in the UK?" question is incorrect. For example:

> Enter if your company is registered in the UK

Should be:

> Select if your company is registered in the UK

Use “Select” rather than “Enter” and “UK” should be uppercase.

gclssvglx:
> On the business details form, the error message for the "Is your company registered in the UK?" question is incorrect. For example:
> 
> > Enter if your company is registered in the UK
> 
> Should be:
> 
> > Select if your company is registered in the UK
> 
> Use “Select” rather than “Enter” and “UK” should be uppercase.

Thanks @simonwhatley good catch.  Messaging and screenshots updated.

app/controllers/coronavirus_form/business_details_controller.rb
```diff
@@ -1,14 +1,16 @@
...
-  TEXT_FIELDS = %w[company_name company_number].freeze
+  REQUIRED_FIELDS = %w[company_name company_is_uk_registered].freeze
+  TEXT_FIELDS = %w[company_name].freeze
 
   def submit
     @form_responses = {
       business_details: sanitized_business_details(params),
     }
 
+    @form_responses[:business_details][:company_number] = "" if @form_responses[:business_details][:company_is_uk_registered] == "No"
```
leenagupte: I think rather than checking against a hard-coded string, it would be safer to check against the value from the locale file.

app/controllers/coronavirus_form/business_details_controller.rb
```diff
@@ -1,14 +1,16 @@
 # frozen_string_literal: true
 
 class CoronavirusForm::BusinessDetailsController < ApplicationController
-  REQUIRED_FIELDS = %w[company_name].freeze
-  TEXT_FIELDS = %w[company_name company_number].freeze
+  REQUIRED_FIELDS = %w[company_name company_is_uk_registered].freeze
+  TEXT_FIELDS = %w[company_name].freeze
```
leenagupte: Why has company_name been removed from this list? Does that mean the length of the value doesn't need to be checked anymore?

app/helpers/check_answers_helper.rb
```diff
@@ -185,7 +185,8 @@ def concat_answer(answer, question)
       joiner = "<br>"
     elsif question.eql?("business_details")
       concatenated_answer << "Company name: #{answer[:company_name]}" if answer[:company_name]
-      concatenated_answer << "Company number: #{answer[:company_number]}" if answer[:company_number]
+      concatenated_answer << "Company registered in the UK: #{answer[:company_is_uk_registered]}" if answer[:company_is_uk_registered]
```
leenagupte: Shouldn't we be showing the answer to this question regardless of where the company is registered?

app/helpers/field_validation_helper.rb
```diff
@@ -102,6 +102,14 @@ def validate_postcode(field, postcode)
     end
   end
 
+  def validate_company_number(field, company_number)
```
leenagupte, line 105: Can you add a test for this method? I know that we've been lax in testing the FieldValidationHelper, but this would be good place to start.

app/views/coronavirus_form/business_details.html.erb
```diff
@@ -22,20 +22,39 @@
 <%= render "govuk_publishing_components/components/input", {
   id: "company_name",
   name: "company_name",
+  heading_size: "s",
```
leenagupte, line 25: Why has this changed?

app/views/coronavirus_form/business_details.html.erb
```diff
@@ -22,20 +22,39 @@
...
+  name: "company_is_uk_registered",
+  error_message: error_items('company_is_uk_registered'),
+  items: t("coronavirus_form.questions.business_details.company_is_uk_registered.options").map.with_index do |(_, item), index|
+    {
+      id: ("company_is_uk_registered" if index == 0),
+      value: item[:label],
+      text: item[:label],
+      checked: @form_responses.dig(:business_details, :company_is_uk_registered) == item[:label],
+      conditional: item[:label] != t("coronavirus_form.questions.business_details.company_is_uk_registered.options.united_kingdom.label") ? nil :
```
leenagupte: I'm not sure how the `conditional` property in this component works, but it seems a bit odd to test the condition is false. Can it be made to be truthy?

config/schemas/form_response.json
```diff
@@ -348,8 +349,11 @@
         "company_name": {
           "type": "string"
         },
+        "company_is_uk_registered": {
+          "$ref": "#/definitions/yes_no"
+        },
         "company_number": {
-          "type": ["string", "null"]
+          "$ref": "#/definitions/company_number"
```
leenagupte: Minor: I don't think a separate definition is necessary as the pattern for company number isn't reusable.

app/views/coronavirus_form/business_details.html.erb
leenagupte: Or better yet, as there are only two items, rather than looping through them, perhaps they could added separately? This would remove the need for conditional logic around the company number input. We do something similar for [product details](https://github.com/alphagov/govuk-coronavirus-business-volunteer-form/blob/master/app/views/coronavirus_form/product_details.html.erb#L81-L114).

leenagupte (commented):
This is looking really good 😄  and I think it's not far off from being ready. The main suggestions and comments have been added inline.

There's a lot going on in this change, so it would be great if some of it could be broken out into separate commits reduce some of the noise. This would have made it much easier to review, and will make it make it easier for future devs to track down the changes.

For example, adding a regex to validate the company number should be in a separate commit. Whilst I think it's good idea, there's no explanation about why we are suddenly doing this as part of this change, or why that particular pattern was chosen.

Another example is adding the new `company_is_uk_registered` field. All of things to add the field, locale entry, new radio button, storing the data etc could be done in one commit, this would make the unfamiliar bit, moving and nesting the company number field, easier to focus on.

Happy to talk about this in person.

app/helpers/check_answers_helper.rb
gclssvglx: Just following the design @leenagupte: 

- With - https://trello-attachments.s3.amazonaws.com/5e8346012feac37ef24b2a9e/5ede05309cbcf53345341ca2/ac1d9610b86c0381bd733842d1b5398b/check-your-answers-001.png
- Without - https://trello-attachments.s3.amazonaws.com/5e8346012feac37ef24b2a9e/5ede05309cbcf53345341ca2/8c0c36fbbf1948979b9d571a0438686a/check-your-answers-002.png

app/helpers/field_validation_helper.rb
gclssvglx: Done - good shout 👍

app/views/coronavirus_form/business_details.html.erb
gclssvglx: It had been missed off - so the text was normal, not <strong>.  Looked odd!

app/controllers/coronavirus_form/business_details_controller.rb
gclssvglx: Done - good spot - thanks @leenagupte 👍

app/controllers/coronavirus_form/business_details_controller.rb
gclssvglx: I've put it back in now as there is no harm in keeping it.  Though, the company number validation does make it un-necessary.  Thanks @leenagupte

app/helpers/check_answers_helper.rb
gclssvglx: Just following the design @leenagupte 🧷

app/views/coronavirus_form/business_details.html.erb
gclssvglx: Was missed, so the text looked odd.  Just making it consistent with all the other input headings on the page.  Thanks @leenagupte

app/views/coronavirus_form/business_details.html.erb
gclssvglx: PENDING: Will take a look at this one...

config/schemas/form_response.json
gclssvglx: Removed - good point @leenagupte - thanks 👍

app/helpers/field_validation_helper.rb
gclssvglx: Done.  Thanks @leenagupte 👍

gclssvglx:
Thanks for the review and the great comments/suggestions @leenagupte.  I've split it down into 3 commits: one to add the company number validation; one to add the new field and ask for the company number if UK registered; and one to fix the missing heading size. 

I tried to get one or two more commits out of it, but they just didn't seem to work as the validation and new field are both atomic pieces of work and can't easily be individually split down.  I do take your point about them being separate though - thanks for pointing that out, think I got lazy!

Thanks 👍

app/helpers/check_answers_helper.rb
leenagupte: The design shows "Company Registered in the UK" for both with and without versions.

gclssvglx: Hi @leenagupte apologies, I thought you were referring to the `company_number` ☹️.  Have updated the code and the tests to reflect this.  Thanks 👍

app/views/coronavirus_form/business_details.html.erb
gclssvglx: Done.  Thanks @leenagupte that looks much cleaner.  Good shout 👍

app/helpers/field_validation_helper.rb
```diff
@@ -106,7 +106,7 @@ def validate_company_number(field, company_number)
     if company_number =~ /^([0-9]{8}|[A-Za-z]{2}[0-9]{6})$/
       []
     else
-      [{ field: field.to_s.sub(".", "_"), text: t("coronavirus_form.errors.company_number_format") }]
+      [{ field: field.to_s.sub(".", "_"), text: t("coronavirus_form.errors.company_number") }]
```
leenagupte, line 109: Nitpick: This should be fixed up to the previous commit as it's just a rename of a field.

config/locales/en.yml
```diff
@@ -626,8 +626,18 @@ en:
...
           custom_error: "Enter the name of your company"
-        company_number:
-          label: "Company number (optional)"
+        company_is_uk_registered:
+          label: "Is your company registered in the UK?"
+          custom_error: "Select if your company is registered in the UK"
+          options:
+            united_kingdom:
+              label: "Yes"
+              input:
```
leenagupte:
```suggestion
              company_number:
```

To make it easier to find and update the content for company number.

spec/helpers/field_validation_helper_spec.rb
```diff
@@ -3,7 +3,7 @@
 RSpec.describe FieldValidationHelper, type: :helper do
   describe "#validate_company_number" do
     let(:field) { "company_number" }
-    let(:error) { [{ field: field, text: t("coronavirus_form.errors.company_number_format") }] }
+    let(:error) { [{ field: field, text: t("coronavirus_form.errors.company_number") }] }
```
leenagupte, line 6: Minor: This should be fixed up too.

leenagupte (approved):
This is looking good! Your tests, as ever, are really great ⭐ .

I've added a couple of minor comments, but nothing blocking. 👍

app/helpers/field_validation_helper.rb
gclssvglx: Fixed.  Thanks @leenagupte 👍

spec/helpers/field_validation_helper_spec.rb
gclssvglx: Fixed.  Thanks @leenagupte 👍

config/locales/en.yml
gclssvglx: Fixed.  Thanks @leenagupte 👍

PR #389, opened 2020-06-18
Add jasmine testing

gclssvglx:
What
----

We need to test that the `product_id` and `reference_number` are redacted and sent in the correct format. 

Currently, if the parameter names changed or the code used to generate the IDs changed, then there would be no test in place to test that redacting the params for GA had broken as a result. 

If the `product_id` and `reference_number` patterns are not matched and redacted, then the params are dropped, so we don't risk sending PII data. The reason that we kept redacted data at all, is that the PAs advise that it is better to have [redacted-number] because it shows that there is missing data and what the redacted data is.

This change adds testing.

Co-authored-by: Ian James <ian.james@digital.cabinet-office.gov.uk>

Links
-----

[Trello cards](https://trello.com/c/M7GUZRpa/585-add-jasmine-testing-to-business-volunteers-form-to-test-redacting-ga-params)

### DIFF
--- a/app/assets/javascripts/stripPII.js
+++ b/app/assets/javascripts/stripPII.js
@@ -1,0 +1,28 @@
+window.GOVUK = window.GOVUK || {}
+
+var stripPII = function (url) {
+  var _splitAndRedact = function (url) {
+    var queryString = url.search.replace(/^\?/,'').split('&')
+    var redactedArray = []
+
+    for (var i = 0; i < queryString.length; i++) {
+      if (queryString[i].startsWith('reference_number') || queryString[i].startsWith('product_id')) {
+        var paramToRedact = queryString[i].split('=')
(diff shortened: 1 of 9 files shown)


app/assets/javascripts/stripPII.js
```diff
@@ -0,0 +1,28 @@
...
+      }
+    }
+
+    return redactedArray.length > 0 ? '?' + redactedArray.join('&') : ''
+  }
+
+  try {
+    return url.protocol + '//' + url.host + url.pathname + _splitAndRedact(url) + url.hash
+  } catch (error) {
+    console.error(error)
```
alex-ju, line 23: If we'd have Sentry setup on the frontend side too, these would be nicely captured – certainly not something for this PR 😄 .

alex-ju (approved):
🎖️

PR #394, opened 2020-06-19
Update check answers page

GDSNewt:
What
----

The 'check your answers' page does not display all questions.  Also, the visual hierarchy of the page can be improved.

This change adds sections to the answers to provide a visual hierarchy.

Screenshots
----

<br>

### Desktop

update-check-your-answers-desktop

<br><br>

### Mobile

update-check-your-answers-mobile

<br><br>

Links
-----

-  [Trello tech card](https://trello.com/c/eGsOEifg/488-update-check-your-answers-page)

- [Trello content design card](https://trello.com/c/BII0T58B/438-how-might-we-improve-the-design-of-the-check-your-answers-page)

Co-authored-by: Graham Lewis <graham.lewis@digital.cabinet-office.gov.uk>


app/views/coronavirus_form/check_answers.html.erb
```diff
@@ -15,21 +15,16 @@
...
-
-<%= render "govuk_publishing_components/components/summary_list", {
-  items: items_part_2,
-} %>
+<% form_outline[:properties].each do |section, property| %>
+  <%= render "govuk_publishing_components/components/heading", {
+    text: section.gsub("-", " ").humanize,
+    margin_bottom: 3,
+  } %>
+  <%= render "govuk_publishing_components/components/summary_list", {
```
alex-ju: Summary list component was updated to allow [custom headings](https://components.publishing.service.gov.uk/component-guide/summary_list/with_custom_section_heading) in preparation for this change. For consistency, we should try to use the `title` and `heading_level`/`heading_size` properties instead of a separate heading component.

config/schemas/form_outline.json
```diff
@@ -0,0 +1,118 @@
...
+        "type": "array",
+        "items": {
+          "type": "string",
+          "enum": [
+            "/contact-details"
+          ]
+        }
+      }
+    }
+  }
```
alex-ju: Minor, we're missing a new blank line at eof.

app/helpers/check_answers_helper.rb
```diff
@@ -202,4 +202,14 @@ def questions
...
     questions["en"]["coronavirus_form"]["questions"].keys
   end
+
+  def base_path(field)
+    field.dig(:edit, :href).split("?").first
+  end
+
+  def question_answers(property, items)
+    items.select do |item|
+      item if property.dig(:items, :enum).include?("/#{base_path(item)}") && item[:value].present?
```
leenagupte: This threw me a bit because I didn't realise for quite a while, that the property being passed in was a schema property, and that why we're checking against the enum.

Same goes for the `base_path` method.

leenagupte (commented):
The page looks good and works as expected, though there is a bit of an issue with the line spacing between the sections which I think Alex's suggestion should sort out.

I'm a bit concerned by the use of a JSON schema, as I think that makes the code harder to follow and the sections harder to update. 

Using a schema seems like a lot for what appears to just need nested arrays of question keys.  I'd consider defining a hash somewhere (in the code in the first instance is fine) that makes it easy to pluck out the keys and use them to look up the section title content from the locale file.

Happy to talk this through in person

app/views/coronavirus_form/check_answers.html.erb
gclssvglx: Thanks @alex-ju 👍  Have updated to use `summary_list` component where possible.

config/schemas/form_outline.json
gclssvglx: Thanks @alex-ju 👍   resolved

app/helpers/check_answers_helper.rb
gclssvglx: Thanks @leenagupte  👍  have changed approach.

gclssvglx:
> The page looks good and works as expected, though there is a bit of an issue with the line spacing between the sections which I think Alex's suggestion should sort out.
> 
> I'm a bit concerned by the use of a JSON schema, as I think that makes the code harder to follow and the sections harder to update.
> 
> Using a schema seems like a lot for what appears to just need nested arrays of question keys. I'd consider defining a hash somewhere (in the code in the first instance is fine) that makes it easy to pluck out the keys and use them to look up the section title content from the locale file.
> 
> Happy to talk this through in person

Hi @leenagupte, @alex-ju thank you very much for the feedback the your suggestions. We've reconsidered our approach to this - which is now based around `partials`.  Would very much appreciate your thoughts and suggestions 👍

app/assets/stylesheets/application.scss
```diff
@@ -45,3 +45,7 @@
     width: 25%;
   }
 }
+
+.gem-c-summary-list {
```
leenagupte, line 50: Why is the fix temporary? Should this change be added to the gem instead?

app/helpers/check_answers_helper.rb
```diff
@@ -1,161 +1,218 @@
...
-    it_services_other
-    it_cost
-  ].freeze
-
-  def items
-    items = questions.map do |question|
-      # We have answers as strings and hashes. The hashes need a little more
-      # work to make them readable.
+  def sections
+    %w[accommodation transport space staff care services other_support location business_details contact_details]
```
leenagupte: Nitpick: This is a very long line, so you might want to consider breaking down into one entry per line.

I see that the entries aren't in alphabetical order. Does that mean the ordering of the entries is important? If yes, can you add a note about that to the commit message.

app/helpers/check_answers_helper.rb
```diff
@@ -1,161 +1,218 @@
...
+  end
 
-      next if skip_question?(question)
+  def match_number_to_list_item(numbers, list_item)
```
leenagupte: What is "number"?

What will this method be used for?

spec/helpers/check_answers_helper_spec.rb
```diff
@@ -1,205 +1,514 @@
...
     end
   end
 
-  describe "#link_to_parent_page" do
-    it "links the user back to the page the question appeared on" do
-      helper.link_to_parent_page("question", "parent_question") do |item|
-        expect(item[:edit][:href]).to include("parent-question?change-answer")
-      end
+  describe "#match_number_to_list_item(numbers, list_item)" do
+    it "returns a matched list item" do
```
leenagupte: I think these test names need a bit more explanation. What is a "matched list item"? What is being matched against what?

app/helpers/check_answers_helper.rb
```diff
@@ -1,161 +1,218 @@
...
-
-    answer.join("<br>")
+  def staff_items
+    [
+      {
+        field: I18n.t("coronavirus_form.check_your_answers.sections.staff.type"),
+        value: session[:offer_staff_type],
+        type: :list_with_numbers,
+        numbers: staff_numbers,
+        number_type: "people",
```
leenagupte: I'm not sure what `list_with_numbers` mean, or why "people" needs to defined as a "number_type". This is the type of change that needs its own commit, so that things like this can be explained.

app/helpers/check_answers_helper.rb
```diff
@@ -1,161 +1,218 @@
...
-      },
-    }]
+  def space_items
+    [
+      {
+        field: I18n.t("coronavirus_form.check_your_answers.sections.space.type"),
+        value: session[:offer_space_type],
+        type: :list_with_numbers,
+        numbers: space_numbers,
+        number_type: "square feet",
```
leenagupte: Again here, I'm not sure why "number_type" is being defined. Is this just content? Can it be added to the locale file instead?

spec/helpers/check_answers_helper_spec.rb
```diff
@@ -1,205 +1,514 @@
 require "spec_helper"
 
 RSpec.describe CheckAnswersHelper, type: :helper do
-  let(:answers_to_skippable_questions) do
+  let(:items) do
```
leenagupte: These items seem very similar to what's returned by `valid_data` in the FormResponseHelper. 
I might be easier to just call that instead.

spec/helpers/check_answers_helper_spec.rb
```diff
@@ -1,205 +1,514 @@
...
-      end
+  describe "#match_number_to_list_item(numbers, list_item)" do
+    it "returns a matched list item" do
+      numbers = { nine_number: 9, ten_number: 10 }
+
+      expect(helper.match_number_to_list_item(numbers, "Ten")).to eq("10")
+    end
+
+    it "returns a matched list item that is not a number" do
+      numbers = { "nine_number" => "9", "ten_number" => "10" }
```
leenagupte: What happens if the number is not a number? 
i.e.

```
numbers = { "nine_number" => "Nine", "ten_number" => "Ten" }
```

spec/helpers/check_answers_helper_spec.rb
```diff
@@ -1,205 +1,514 @@
...
+        .to include("fred@bedrock.com")
     end
   end
 
-  describe "#link_to_parent_page" do
-    it "links the user back to the page the question appeared on" do
-      helper.link_to_parent_page("question", "parent_question") do |item|
-        expect(item[:edit][:href]).to include("parent-question?change-answer")
-      end
+  describe "#match_number_to_list_item(numbers, list_item)" do
```
leenagupte: I'm not sure what this method is supposed to be returning. Is it the highest number in the hash? Do the key names matter?

app/views/coronavirus_form/check_your_answers/helpers/_list_with_numbers.html.erb
```diff
@@ -0,0 +1,5 @@
+<ul class="govuk-list govuk-list--bullet">
+  <% items.each do |item| %>
+    <li><%= item %> (<%= match_number_to_list_item(numbers, item) %> <%= number_type %>)</li>
```
leenagupte: I'm not quite sure what this line looks like. A view test could make that clearer.

app/views/coronavirus_form/check_your_answers/helpers/_question_answers.html.erb
```diff
@@ -0,0 +1,18 @@
+<%= render "govuk_publishing_components/components/summary_list", {
+  title: question_title,
```
leenagupte: I wonder if this should be called `section_title` as I think that's what it is?

app/views/coronavirus_form/check_your_answers/helpers/_question_answers.html.erb
```diff
@@ -0,0 +1,18 @@
...
+  heading_level: 2,
+  heading_size: "l",
+  items: question_items,
+  edit: {
+    href: "#{question_href}?change-answer",
+  },
+} %>
+<%= render "govuk_publishing_components/components/summary_list", {
+  title: detail_title,
```
leenagupte: I think this is really the question title, right?

app/views/coronavirus_form/check_your_answers/helpers/_summary_list_multiple_items.html.erb
```diff
@@ -0,0 +1,46 @@
...
+    Change<span class="govuk-visually-hidden"></span></a></li>
+  </ul>
+  <dl class="govuk-summary-list">
+    <% detail_items.each do |item| %>
+      <div class="govuk-summary-list__row">
+        <dt class="govuk-summary-list__key">
+          <%= item[:field] %>
+        </dt>
+        <dd class="govuk-summary-list__value">
+          <% if item[:type] == :list_with_numbers %>
```
leenagupte: I'm not really sure what the difference between a question and a detail is, so that may affect the answer, but can this two summary lists be combined? The only differences I can see between them are the "items"being looped through, and this extra conditional that checks `list_with_numbers`.

leenagupte: I'm also wondering if the changes being added here, i.e. displaying items differently depending on the object type, should be added to the main summary-list component as it feels like we're essentially forking that here.

app/views/coronavirus_form/check_your_answers/sections/_location.html.erb
```diff
@@ -0,0 +1,19 @@
+<%= render "govuk_publishing_components/components/summary_list", {
+  title: t("coronavirus_form.check_your_answers.sections.location.title"),
+  heading_level: 2,
+  heading_size: "l",
+  edit: {
+    href: "location?change-answer",
+  }
+} do %>
+  <dl class="govuk-summary-list">
```
leenagupte: Is this a summary-list that's being passed as a block to the main summary-list component? Is that because the main summary-list doesn't allow for descriptions. I wondering if it's worth adding this to the main component? @injms will able to answer that better than me.

app/views/coronavirus_form/check_your_answers/sections/_services.html.erb
```diff
@@ -0,0 +1,46 @@
+<%= render "coronavirus_form/check_your_answers/helpers/summary_list_multiple_items",
+  question_title: t("coronavirus_form.check_your_answers.sections.services.title"),
```
leenagupte: I've just been confused by what "services" are under "sections". I thought it was a list of sub-sections! We might want to consider renaming this to "services_or_expertise"

app/views/coronavirus_form/check_your_answers/sections/_services.html.erb
```diff
@@ -0,0 +1,46 @@
...
+<div class="govuk-!-margin-bottom-9">
+  <%= render "govuk_publishing_components/components/summary_list", {
+    title: t("coronavirus_form.check_your_answers.sections.services.it.title"),
+    heading_level: 3,
+    heading_size: "m",
+    edit: {
+      href: "it-services?change-answer",
+    }
+  } do %>
+    <dl class="govuk-summary-list">
```
leenagupte: Again this feels like we're building another summary-list inside a summary-list. Why is that? Can the main component be updated to accommodate this need?

app/views/coronavirus_form/check_your_answers/sections/_care.html.erb
```diff
@@ -0,0 +1,8 @@
+<%= render "coronavirus_form/check_your_answers/helpers/summary_list_multiple_items",
```
leenagupte: It feels like this would be a good place to call the summary list twice, rather than doing it in a partial.

app/views/coronavirus_form/check_your_answers/helpers/_summary_list_multiple_items.html.erb
leenagupte: It would be great if you can add your reasoning behind why this partial has been created and the other implementations you considered ( i.e. why isn't the summary-list component being used?) to the commit message.

app/views/coronavirus_form/check_your_answers/sections/_accommodation.html.erb
```diff
@@ -0,0 +1,8 @@
+<%= render "coronavirus_form/check_your_answers/helpers/question_answers",
+  question_title: t("coronavirus_form.check_your_answers.sections.accommodation.title"),
+  question_items: question_items("accommodation"),
+  question_href: "accommodation",
+  detail_title: t("coronavirus_form.check_your_answers.details"),
+  detail_items: accommodation_items,
+  detail_href: "rooms-number"
```
leenagupte: I wonder if this hash could be returned by `accommodation_items` in the helper, rather than being created in the view? The question applies to the hashes being passed into the other partials too.

leenagupte (commented):
Overall this works well. And as ever, your tests are fantastic! 🎉  

I think there's a few things left to do to make it easier to add and re-jig the sections in the future.

I've added inline comments with some questions and recommendation. 

My main concern is about how much is going on in the views. 
I had assumed it would be a case of calling the summary-list component at most twice for each section, once for the guard question, and the other for rest of the questions, but instead it feels like we're recreating the summary-list component in some of the views.

I think the complexity should be moved back the helper where possible.
For example, the code that generates this line:

```
Cleaners (342 people)
```

Is split between the helper and the view. I think it would helpful to move it all into the helper and test it there.

Finally, it would have really helped understand what was going on if the changes were committed by section. For example, if the section text, view, helper method and tests were in the same commit with a description of why some choices were made.

Happy to talk this through in person.

app/views/coronavirus_form/check_your_answers/helpers/_summary_list_multiple_items.html.erb
```diff
@@ -0,0 +1,46 @@
+<div class="gem-c-summary-list ">
+  <h2 class="govuk-heading-l gem-c-summary-list__group-title"><%= question_title %></h2>
+  <ul class="govuk-summary-list__actions-list gem-c-summary-list__group-actions-list">
+    <li class="govuk-summary-list__actions-list-item"> <a class="govuk-link" href="<%= question_href %>?change-answer">
+    Change<span class="govuk-visually-hidden"></span></a></li>
```
injms: This `<span class="govuk-visually-hidden"></span>` isn't doing anything - so we can either remove it or add the question in here so it's read aloud by a screen reader.

gclssvglx: Thanks @injms 👍

app/views/coronavirus_form/check_your_answers/helpers/_summary_list_multiple_items.html.erb
```diff
@@ -0,0 +1,46 @@
...
+            <%= item[:value] %>
+          <% end %>
+        </dd>
+      </div>
+    <% end %>
+  </dl>
+  <h3 class="govuk-heading-m gem-c-summary-list__group-title"><%= detail_title %></h3>
+  <ul class="govuk-summary-list__actions-list gem-c-summary-list__group-actions-list">
+    <li class="govuk-summary-list__actions-list-item"> <a class="govuk-link" href="<%= detail_href %>?change-answer">
+    Change<span class="govuk-visually-hidden"></span></a></li>
```
injms: Here's another `span` that needs either text or to be removed.

injms: Would using the `link_to` method make this link easier to read? 

We should also indent it to help make the elements more readable:
```suggestion
    <li class="govuk-summary-list__actions-list-item"> 
      <a class="govuk-link" href="<%= detail_href %>?change-answer">
        Change<span class="govuk-visually-hidden"></span>
      </a>
    </li>
```

app/views/coronavirus_form/check_your_answers/sections/_location.html.erb
injms: Agreed, it could be worth adding descriptions to the component - but from just looking at this I don't think that this adds anything extra that isn't already in the summary list component. I'll have a dig and see.

app/helpers/check_answers_helper.rb
```diff
@@ -1,161 +1,218 @@
...
+      cleaners_number: session[:offer_staff_number][:cleaners_number],
+      developers_number: session[:offer_staff_number][:developers_number],
+      medical_staff_number: session[:offer_staff_number][:medical_staff_number],
+      office_staff_number: session[:offer_staff_number][:office_staff_number],
+      security_staff_number: session[:offer_staff_number][:security_staff_number],
+      trainers_or_coaches_number: session[:offer_staff_number][:trainers_number],
+      translators_number: session[:offer_staff_number][:translators_number],
+      other_staff_number: session[:offer_staff_number][:other_staff_number],
+    }.with_indifferent_access
+  end
```
reggieb: I think this has the same effect, and makes it a little clear what the difference is between the two hashes:
```ruby
def staff_numbers
  staff_numbers = session[:offer_staff_number]
  staff_numbers[:trainers_or_coaches_number] = staff_numbers.delete(:trainers_number)
  staff_numbers.with_indifferent_access
end
```

app/helpers/check_answers_helper.rb
```diff
@@ -1,161 +1,218 @@
...
+        field: I18n.t("coronavirus_form.questions.#{question}.title"),
+        value: session[question],
+        type: :list,
+      }]
+    else
+      [{
+        field: I18n.t("coronavirus_form.questions.#{question}.title"),
+        value: session[question],
       }]
     end
```
reggieb, line 24: I'd suggest:
```ruby
def question_items(question)
  question_items = {
    field: I18n.t("coronavirus_form.questions.#{question}.title"),
    value: session[question],
  }
  question_items[:type] = :list if session[question].is_a?(Array)
  [question_items]
end
```
Again - that makes it clear what difference the question being in a array makes to the output

app/helpers/check_answers_helper.rb
```diff
@@ -1,161 +1,218 @@
...
+      },
+      {
+        field: I18n.t("coronavirus_form.check_your_answers.sections.accommodation.description"),
+        value: session[:accommodation_description],
+      },
+      {
+        field: I18n.t("coronavirus_form.check_your_answers.charge"),
+        value: session[:accommodation_cost],
+      },
+    ]
```
reggieb, line 56: I think an I18n scope would be useful here to reduce the length of some of the translations paths. Also, `I18n.t("coronavirus_form.check_your_answers.charge")` is used repeatedly so could be defined separately. So for example, if these two methods were added to the file:
```ruby
def sections_scope
  "coronavirus_form.check_your_answers.sections"
end

def charge_text
  I18n.t("coronavirus_form.check_your_answers.charge")
end
```
Then this method could be rewritten as:
```ruby
def accommodation_items
    [
      {
        field: I18n.t("accommodation.type", scope: sections_scope),
        value: session[:rooms_number],
      },
      {
        field: I18n.t("accommodation.description", scope: sections_scope),
        value: session[:accommodation_description],
      },
      {
        field: charge_text,
        value: session[:accommodation_cost],
      },
    ]
end
```
And the other methods in this file could similarly be simplified.

config/locales/en.yml
```diff
@@ -230,6 +230,59 @@ en:
       link:
         label: "Offer tests or testing facilities (opens in a new window)"
       external_link: "https://www.gov.uk/guidance/help-the-government-increase-coronavirus-covid-19-testing-capacity"
+    check_your_answers:
+      charge: "How much you will charge"
```
reggieb, line 204: The `"` marks are redundant. YAML's default behaviour is to convert the entries to strings. Quotes are only needed if the text contains characters that the YAML parser may misinterpret (such as `:`). So:
```yaml
charge: How much you will charge
```

reggieb (commented):
Some small suggestions that can be ignored.

app/helpers/check_answers_helper.rb
leenagupte: I didn't know you could use scope like that. Nice!

app/assets/stylesheets/application.scss
GDSNewt: A more permanent solution will be incorporated when the summary list component is updated to handle lists. I think @injms was going to have a look at that.

app/helpers/check_answers_helper.rb
GDSNewt: Thanks for this suggestion, we turned this into a hash with keys in numerical order.

app/helpers/check_answers_helper.rb
GDSNewt: This is now outdated 😺

spec/helpers/check_answers_helper_spec.rb
GDSNewt: This is outdated 😺

app/helpers/check_answers_helper.rb
GDSNewt: This is outdated 😺

app/views/coronavirus_form/check_your_answers/sections/_location.html.erb
GDSNewt: This is outdated 😺

app/views/coronavirus_form/check_your_answers/sections/_services.html.erb
GDSNewt: This is outdated 😺

app/views/coronavirus_form/check_your_answers/sections/_services.html.erb
GDSNewt: This is outdated 😺

app/views/coronavirus_form/check_your_answers/sections/_care.html.erb
GDSNewt: This is outdated 😺

app/views/coronavirus_form/check_your_answers/sections/_accommodation.html.erb
GDSNewt: This is outdated 😺

app/views/coronavirus_form/check_your_answers/helpers/_summary_list_multiple_items.html.erb
GDSNewt: This is outdated 😺

app/views/coronavirus_form/check_your_answers/helpers/_summary_list_multiple_items.html.erb
GDSNewt: This is outdated 😺

GDSNewt: This is outdated 😺

app/helpers/check_answers_helper.rb
GDSNewt: Thanks Rob, we went with this suggestion 👍

app/helpers/check_answers_helper.rb
GDSNewt: Thanks Rob, we went with this suggestion 👍

app/helpers/check_answers_helper.rb
GDSNewt: This is outdated 😺

config/locales/en.yml
GDSNewt: Thanks Rob, yeah that makes total sense. That said, I think this should probably be a separate card, and be done across all three forms.

app/views/coronavirus_form/check_your_answers/helpers/_list.html.erb
```diff
@@ -0,0 +1,7 @@
+<% if items.any? %>
+  <ul class="govuk-list govuk-list--bullet">
+    <% items.each do |item| %>
+      <li><%= item %></li>
+    <% end %>
+  </ul>
+<% end %>
```
reggieb: There are a few files like this that don't have a blank line at the end. That can cause problems sometimes.

https://stackoverflow.com/questions/2287967/why-is-it-recommended-to-have-empty-line-in-the-end-of-a-source-file

app/helpers/check_answers_helper.rb
```diff
@@ -1,154 +1,255 @@
...
+      2 => "transport",
+      3 => "space",
+      4 => "staff",
+      5 => "care",
+      6 => "services",
+      7 => "other_support",
+      8 => "location",
+      9 => "business_details",
+      10 => "contact_details",
+    }.sort.to_h.values
```
reggieb: In Ruby (I think post version 2.1) hashes and arrays are ordered, so they return in the same order consistently. Also the `sort` is acting on the keys so sorting doesn't change the order. That means that the content before the `.sort.to_h` is exactly the same as afterwards. Also I'm not sure what value the numbers provided. I'd suggest simplifying to

```ruby
def sections
  %w[
    accommodation
    transport
    space
    staff
    care
    services
    other_support
    location
    business_details
    contact_details
  ]
end
```

app/helpers/check_answers_helper.rb
```diff
@@ -1,154 +1,255 @@
...
-      value: sanitize(transport_type_info),
-      edit: {
-        href: "transport-type?change-answer",
+  def create_list_with_named_values(names, counts, postfix)
+    return "" if names.blank?
+
+    list = []
+    names.each do |name|
+      list << "#{name} (#{humanize_count(counts[counts_key(name)], postfix)})"
+    end
```
reggieb: I'd suggest using `map` here, rather than creating an empty array and pushing into it within an `each` iteration.
```ruby
list = names.map do |name|
  "#{name} (#{humanize_count(counts[counts_key(name)], postfix)})"
end
```

config/initializers/inflections.rb
```diff
@@ -12,6 +13,6 @@
 # end
 
 # These inflection rules are supported but not enabled by default:
-# ActiveSupport::Inflector.inflections(:en) do |inflect|
-#   inflect.acronym 'RESTful'
-# end
+ActiveSupport::Inflector.inflections(:en) do |inflect|
+  inflect.irregular "square foot", "square feet"
```
reggieb: I think the string that needs correcting is "foot", which appears to pluralize as "foots". It being preceded by "square" is irrelevant. So:
```ruby
inflect.irregular "foot", "feet"
```

app/helpers/check_answers_helper.rb
gclssvglx: Thanks @reggieb 👍  Happy to revert this change back to `%w[...]` as you suggest.  However, the suggestion that caused us to change this was that using an array did not sufficiently indicate that we are expecting an explicit - and predictable - order.  Personally, I feel that - as code is also documentation - the `{ 1 => '', 2 =>, ... x => }sort.to_h.values` approach (whilst redundant) does precisely describe our intent and the desired behaviour.

config/initializers/inflections.rb
gclssvglx: Thanks @reggieb 👍  we'll take another look at this.  However, I'm sure when we tested this as 'foot' and 'feet' we kept getting 'foots'.  But, with 'square foot' and 'square feet' it worked as expected.

app/helpers/check_answers_helper.rb
gclssvglx: Thanks @reggieb 👍  Good spot - we will fix this.

app/assets/stylesheets/application.scss
leenagupte: Is there a PR in progress in the publishing components gem?  I'm concerned that it will forgotten about?

app/views/coronavirus_form/check_answers.html.erb
```diff
@@ -19,13 +19,8 @@
...
 } %>
 
-<%= render "govuk_publishing_components/components/heading", {
-  text: t("check_your_answers.heading"),
-  heading_level: 2,
-  margin_bottom: 3,
-} %>
-
-<%= tag.p t("check_your_answers.confirmation"), class: "govuk-body" %>
+<%= render "govuk_publishing_components/components/inset_text", {
```
leenagupte, line 22: This is much nicer!

app/helpers/check_answers_helper.rb
leenagupte: I agree with Rob. I think updating this hash could become frustrating if everything else needs to be re-numbered.

There are two ways to communicate the implicit ordering.
- Rename the method to something like `ordered_sections`
- Add a note to the commit message about the ordering.

spec/helpers/check_answers_helper_spec.rb
leenagupte: @gclssvglx @GDSNewt Did you decide against using `valid_data`? Can you tell me why?

config/initializers/inflections.rb
gclssvglx: Looks like you are correct @reggieb 👍  I must have goofed somehow!  We will fix this.

app/views/coronavirus_form/check_your_answers/helpers/_list.html.erb
gclssvglx: Thanks @reggieb 👍  Good spot - we will fix this.

app/helpers/check_answers_helper.rb
```diff
@@ -1,154 +1,47 @@
...
-
-  def link_to_parent_page(question, parent_question)
-    [{
-      field: t("coronavirus_form.questions.#{question}.title"),
-      value: sanitize(concat_answer(session[question.to_sym], question)),
-      edit: {
-        href: "#{parent_question.dasherize}?change-answer",
+      {
+        field: I18n.t("coronavirus_form.check_your_answers.sections.accommodation.description"),
+        value: session[:accommodation_description],
```
leenagupte, line 50: The edit links are no longer being added here. Can you add a note to the commit message explaining why they are no longer being generated here, and what the plan is for them?

leenagupte (commented):
It looks like the PR is in the midst of being updated, so I'm going to delay the full review. I got about 3 commit into the review. Comments so far have been added inline.

app/helpers/check_answers_helper.rb
reggieb: I think the main original criticism of the array was that it was all on one line. Making it multi-line makes it easier to see what the order is and manipulate it as needed. With the numbered array, if you want to change the order, you'll need to move the elements and update all the numbers. So at least twice the complexity.

Personally - of Leena's options, I think the comment would be my favourite.

gclssvglx: Thanks @leenagupte, @reggieb - I think we're going to have to agree to disagree on this one and I'm clearly out-gunned.  So, I've updated as requested and added a commit comment.

app/views/coronavirus_form/check_answers.html.erb
gclssvglx: 👍

app/helpers/check_answers_helper.rb
gclssvglx: Done - comment on appropriate commit.  Thanks 👍

app/assets/stylesheets/application.scss
gclssvglx: We'll look into it - or create one!  Thanks 👍

app/helpers/check_answers_helper.rb
```diff
@@ -1,154 +1,254 @@
...
+      field: t("location.description", scope: sections_scope),
+      value: render("coronavirus_form/check_your_answers/helpers/list", items: session["location"]),
+    }]
+  end
+
+  def business_detail_items
+    [
+      {
+        field: I18n.t("business_details.name", scope: sections_scope),
+        value: session[:business_details][:company_name],
```
reggieb: I'd suggest using `dig` for getting values within a nested hash, as it handles the absence of the higher level elements more neatly. So here for example:
```ruby
value: session.dig(:business_details, :company_name),
```
I've highlighted this occurrence, but the same pattern appears a lot in this PR.

app/helpers/check_answers_helper.rb
```diff
@@ -1,154 +1,254 @@
...
+      space
+      staff
+      care
+      services
+      other_support
+      location
+      business_details
+      contact_details
+    ]
+  end
```
reggieb, line 4: I thought there was going to be a comment added to describe the significance of the order.

app/helpers/check_answers_helper.rb
```diff
@@ -1,154 +1,254 @@
...
+      },
+      {
+        field: I18n.t("business_details.location", scope: sections_scope),
+        value: business_detail_location,
+      },
+    ]
+  end
+
+  def business_detail_location
+    if session[:business_details][:company_postcode].present?
```
reggieb: Definitely use `dig` here where checking for presence:
```ruby
if session.dig(:business_details, :company_postcode).present?
```

app/views/coronavirus_form/check_answers.html.erb
```diff
@@ -13,20 +13,16 @@
...
+  <% sections.each do |section| %>
+    <%= render "coronavirus_form/check_your_answers/sections/#{section}" %>
+  <% end %>
+</div>
 
-<%= render "govuk_publishing_components/components/summary_list", {
-  items: items,
+<%= render "govuk_publishing_components/components/inset_text", {
+  text: t("check_your_answers.confirmation")
 } %>
```
reggieb, line 24: Really picky, but the curly braces are not needed. So:
```ruby
render "govuk_publishing_components/components/inset_text", text: t("check_your_answers.confirmation")
```

app/views/coronavirus_form/check_your_answers/sections/_accommodation.html.erb
```diff
@@ -0,0 +1,19 @@
+<%= render "govuk_publishing_components/components/summary_list", {
+  title: t("coronavirus_form.check_your_answers.sections.accommodation.title"),
+  heading_level: 2,
+  heading_size: "l",
+  items: question_items("accommodation"),
+  edit: {
+    href: "accommodation?change-answer",
+  },
+} %>
```
reggieb, line 9: Again the outer curly brace is not needed

reggieb (approved):
I've added some really minor suggestions, but am happy to approve as is. However, I think @leenagupte should have a chance to review before this is merged.

leenagupte (dismissed):
Before looking at the code changes, I ran through the form a few times to make sure that the flow works as expected. Unfortunately, I've found a couple of bugs.

I don't know if this is an existing bug, but the check answers page does not update correctly if the user changes their answer from "yes" to "no" for any of the "Can you ...." questions.

For example, in the "accommodation" question. If the user first answers "yes", fills in the details, then gets to the check answers page, and changes their answer to "no", the previously filled in accommodation questions are still shown, with the exception of the "how much charge" question, which is still displayed, but the answer has been blanked out (though this is as expected).

I can't tell from the design what is supposed to happen in this situation. i.e. whether the details section is still supposed to be displayed, but be blank, or whether it should be completely hidden.

The user can still change the details of the question to which they've answered "No", but they can't remove the details completely as the form validation doesn't let them save and continue on the journey.

Also, if you change the "Services or Expertise" answer to "I don't have any services or expertise", the check answers helper throws the following error:

```
ActionView::Template::Error at /check-your-answers
undefined method `any?' for nil:NilClass
```

The error appears to be occurring here:

```
<% if items.any? %>
  <ul class="govuk-list govuk-list--bullet">
    <% items.each do |item| %>
      <li><%= item %></li>
    <% end %>
  </ul>
```

I think this error occurs because I'd previously selected "IT services" and "Construction Services". As long as one of these remains selected, the form still works, but if one of them is both removed, the error occurs.


I cleared my session and tried the form again, and selected "No" or "no expertise" for each answer the first time around, and still got the error above. i.e.

```
ActionView::Template::Error at /check-your-answers
undefined method `any?' for nil:NilClass
```

```
<% if items.any? %>
  <ul class="govuk-list govuk-list--bullet">
    <% items.each do |item| %>
      <li><%= item %></li>
    <% end %>
  </ul>
```

I've played around with a few different form combinations, but I can't tell where in the code it's breaking.

I think it's best to give you time to investigate these bugs before I re-review the code as I'm sure some tweaks will be needed.

app/helpers/check_answers_helper.rb
gclssvglx: Thanks @reggieb - yes, it's in this commit https://github.com/alphagov/govuk-coronavirus-business-volunteer-form/pull/394/commits/3a2bfbdfa08e2444eac2e0237520aa507b1a547b

app/views/coronavirus_form/check_answers.html.erb
gclssvglx: Hi @reggieb, sorry you are incorrect - the braces are required.  The page throws a syntax error without them.

app/views/coronavirus_form/check_your_answers/sections/_accommodation.html.erb
gclssvglx: Hi @reggieb, sorry you are incorrect - the braces are required.  The page throws a syntax error without them.

reggieb: Fair enough!

spec/helpers/check_answers_helper_spec.rb
```diff
@@ -1,203 +1,61 @@
...
-        expect(item[:edit][:href]).to include("?change-answer")
-      end
+      expect(helper.question_items("location")).to eq(
+        [
+          {
+            field: I18n.t("coronavirus_form.questions.location.title"),
+            value: render("coronavirus_form/check_your_answers/helpers/list", items: form_data[:location]),
+          },
+        ],
+      )
```
leenagupte: I think this might be in the wrong commit?

app/helpers/check_answers_helper.rb
```diff
@@ -1,154 +1,47 @@
...
+      business_details
+      contact_details
+    ]
+  end
+
+  def question_items(question)
+    if session[question].is_a?(Array)
+      [{
+        field: I18n.t("coronavirus_form.questions.#{question}.title"),
+        value: render("coronavirus_form/check_your_answers/helpers/list", items: session[question]),
```
leenagupte: I think the partial referenced here is being referenced in a later commit? This is the partial that is throwing the error I mentioned in my "form flow" review, so I'm guessing it's either going to change slightly, or be replaced with: https://github.com/alphagov/govuk_publishing_components/pull/1595

app/views/coronavirus_form/check_your_answers/helpers/_list.html.erb
```diff
@@ -0,0 +1,7 @@
+<% if items.any? %>
```
leenagupte: Ah, hello little offender! I think this is the partial that should be with the transport changes.
Hopefully the use of it will soon be able to be replace with https://github.com/alphagov/govuk_publishing_components/pull/1595.

app/helpers/check_answers_helper.rb
gclssvglx: Good spot @reggieb 👍

app/helpers/check_answers_helper.rb
gclssvglx: Thanks @reggieb 👍

gclssvglx:
https://github.com/alphagov/govuk-coronavirus-business-volunteer-form/pull/394#discussion_r451408852 - no, we did go with your suggestion.  Thanks 👍

app/helpers/check_answers_helper.rb
```diff
@@ -1,154 +1,56 @@
...
+    ]
+  end
+
+  def no_answer?(section)
+    answer = session[section]
+    no_answer_options = [
+      I18n.t("coronavirus_form.questions.accommodation.options.no_option.label"),
+    ].sort.uniq
+
+    answer.is_a?(Array) ? no_answer_options.include?(answer.first) : no_answer_options.include?(answer)
```
leenagupte: This seems quite complicated. Is this check happening because a question could have more than one "no answer"?

What happens if the answer for the section happens to match a "no answer" for another section, but not the "no answer" for its section?

I appreciate that you added this method to the helper to make it testable, and as I much as I hate logic in views, I think that might have been easier.

app/views/coronavirus_form/check_your_answers/sections/_accommodation.html.erb
```diff
@@ -0,0 +1,21 @@
...
+
+<% unless no_answer?(:accommodation) %>
```
leenagupte: As mentioned earlier, it might have been easier if this conditional did the work, i.e. rather than call the method:

```
unless session[:accommodation] == I18n.t("coronavirus_form.questions.accommodation.options.no_option.label")
```

app/helpers/check_answers_helper.rb
```diff
@@ -1,154 +1,56 @@
...
+      care
+      services
+      other_support
+      location
+      business_details
+      contact_details
+    ]
+  end
+
+  def no_answer?(section)
```
leenagupte: Nitpick: The name "no_answer?" implies that the we're checking if the user answered _at all_, not whether they selected "no". Perhaps it could renamed to `show_section_details?`

spec/helpers/check_answers_helper_spec.rb
```diff
@@ -1,203 +1,87 @@
...
 
-      expect(helper.transport_type_info).to eq(expected_answer)
+      expect(helper.no_answer?(:accommodation)).to be true
     end
 
-    it "only concatenates the fields that have a value" do
-      session["transport_type"] = [
-        I18n.t("coronavirus_form.questions.transport_type.options.moving_people.label"),
-      ]
+    it "returns true if 'I cannot offer expertise'" do
```
leenagupte: Is this in the correct commit?

spec/helpers/check_answers_helper_spec.rb
```diff
@@ -1,203 +1,87 @@
...
 
-  describe "#transport_type_info" do
-    it "concates transport type and description" do
-      session["transport_type"] = [
-        I18n.t("coronavirus_form.questions.transport_type.options.moving_people.label"),
-        I18n.t("coronavirus_form.questions.transport_type.options.moving_goods.label"),
-        I18n.t("coronavirus_form.questions.transport_type.options.other.label"),
-      ]
+  describe "#no_answer?" do
+    it "returns false if Yes" do
```
leenagupte:
```suggestion
    it "returns false if user answers Yes to offering accommodation" do
```

spec/helpers/check_answers_helper_spec.rb
```diff
@@ -1,203 +1,87 @@
...
 
-      session["transport_description"] = "Transport description"
+      expect(helper.no_answer?(:accommodation)).to be false
+    end
 
-      expected_answer = "#{I18n.t('coronavirus_form.questions.transport_type.options.moving_people.label')}, " \
-                          "#{I18n.t('coronavirus_form.questions.transport_type.options.moving_goods.label')}, and " \
-                          "#{I18n.t('coronavirus_form.questions.transport_type.options.other.label')}<br>" \
-                          "#{session['transport_description']}"
+    it "returns true if No" do
```
leenagupte:
```suggestion
    it "returns true if user answers No to offering accommodation do
```

spec/helpers/check_answers_helper_spec.rb
```diff
@@ -1,203 +1,87 @@
...
+      expect(helper.no_answer?(:accommodation)).to be true
     end
-  end
 
-  describe "#offer_care_qualifications" do
-    it "adds a query string to the link for each item" do
-      helper.offer_care_qualifications.each do |item|
-        expect(item[:edit][:href]).to include("?change-answer")
-      end
+    it "returns false if anything else" do
```
leenagupte:
```suggestion
    it "returns false if users answers anything but No to offering accommodation" do
```

app/views/coronavirus_form/check_your_answers/helpers/_list.html.erb
```diff
@@ -0,0 +1,7 @@
+<% if items && items.any? %>
```
leenagupte: Ah. I think you should be in the commit for the accommodation section.

app/helpers/check_answers_helper.rb
```diff
@@ -18,6 +18,7 @@ def no_answer?(section)
     answer = session[section]
     no_answer_options = [
       I18n.t("coronavirus_form.questions.accommodation.options.no_option.label"),
+      I18n.t("coronavirus_form.questions.offer_transport.options.option_no.label"),
```
leenagupte: Should there be tests for this?

app/helpers/check_answers_helper.rb
```diff
@@ -38,6 +39,25 @@ def question_items(question)
...
+  def counts_key(name)
+    "#{name.downcase.gsub(' ', '_')}_number".to_sym
+  end
+
+  def humanize_count(count, postfix)
+    number = number_with_delimiter(count, delimiter: ",")
+    pluralize(number, postfix)
+  end
+
+  def create_list_with_named_values(names, counts, postfix)
```
leenagupte:
```suggestion
  def create_list_with_named_values(items, counts, postfix)
```
This is not a blocker, but `names` brings to mind "first name", "last name" etc, rather than being generic as I think this method is trying to be.

app/helpers/check_answers_helper.rb
```diff
@@ -38,6 +39,25 @@ def question_items(question)
...
+    return "" if names.blank?
+
+    list = names.map do |name|
+      "#{name} (#{humanize_count(counts[counts_key(name)], postfix)})"
```
leenagupte: Minor: `counts[counts_key(name)]` is a bit hard to read. I'd consider moving it to its own line and assigning it to a well named variable.

app/helpers/check_answers_helper.rb
```diff
@@ -38,6 +39,25 @@ def question_items(question)
     end
   end
 
+  def counts_key(name)
+    "#{name.downcase.gsub(' ', '_')}_number".to_sym
```
leenagupte: It feels like this needs to match something, but it's not clear what. I think this rule could be easily broken by someone new to the code. Is this method needed or can the caller of `create_list_with_named_values` tell it which numbers method to use?

config/initializers/inflections.rb
```diff
@@ -12,6 +13,6 @@
 # end
 
 # These inflection rules are supported but not enabled by default:
-# ActiveSupport::Inflector.inflections(:en) do |inflect|
-#   inflect.acronym 'RESTful'
-# end
+ActiveSupport::Inflector.inflections(:en) do |inflect|
+  inflect.irregular "foot", "feet"
```
leenagupte, line 6: It's so weird that this is needed. You'd think "feet" would be included by default.

spec/helpers/check_answers_helper_spec.rb
```diff
@@ -69,6 +73,77 @@
     end
   end
 
+  describe "#counts_key" do
```
leenagupte: Great tests, as always.

spec/helpers/check_answers_helper_spec.rb
```diff
@@ -69,6 +73,77 @@
...
+
+    it "returns an empty array given an empty array" do
+      expect(helper.create_list_with_named_values([], {}, ""))
+        .to eq(render("coronavirus_form/check_your_answers/helpers/list", items: []))
+    end
+
+    it "returns a valid html list" do
+      names = ["Other", "Warehouse space", "Office space"]
+      counts = { other_number: 10_000, warehouse_space_number: 1, office_space_number: 300 }
+      postfix = "person"
```
leenagupte: Nitpick: It's odd that the test is combining space with people rather than an area measurement. It gets the point across, so it's not a blocker.

spec/helpers/check_answers_helper_spec.rb
```diff
@@ -469,4 +473,64 @@
...
+
+      expect(helper.business_detail_items.pluck(:field))
+        .to include(I18n.t("coronavirus_form.check_your_answers.sections.business_details.size"))
+      expect(helper.business_detail_items.pluck(:value))
+        .to include(form_data.dig(:business_details, :company_size))
+    end
+
+    it "contains a location field where postcode is an empty string" do
+      data_copy = form_data
+      data_copy[:business_details][:company_postcode] = ""
```
leenagupte:
```suggestion
      data_copy = form_data.tap do |form_data|
        form_data[:business_details].delete(:company_postcode)
      end
```

This is a minor style suggestion, and not a blocker.

leenagupte (commented):
Thanks for fixing the bugs! It looks like the problem with the list partial has been solved 🎉 

As always your tests are great and really help clarify what's going on in the code. The use of the main summary_list component is great too! It makes the new partials much easier to work with.

I've added some inline comments. Most of which are minor. The only lingering concern I have is with the `no_answer?` method and its surrounding logic. I've commented about this inline, but I think it's little hard to read which makes what it's doing a bit of a head scratcher. I think it could also do with some more tests. For example, what happens if a user answers "No" to a question that isn't specifically included in the no_answer list, but has text that matches the "no answer" for another question?

On the whole though it's looking good! It's almost there!

spec/helpers/check_answers_helper_spec.rb
gclssvglx: Thanks @leenagupte 👍  Outdated - resolved.

app/helpers/check_answers_helper.rb
gclssvglx: Thanks @leenagupte 👍  Outdated - resolved.

app/views/coronavirus_form/check_your_answers/helpers/_list.html.erb
gclssvglx: Thanks @leenagupte 👍  Outdated - resolved.

app/helpers/check_answers_helper.rb
gclssvglx: Thanks @leenagupte 👍  Outdated - resolved.

app/views/coronavirus_form/check_your_answers/sections/_accommodation.html.erb
gclssvglx: Thanks @leenagupte 👍  Outdated - resolved.

app/helpers/check_answers_helper.rb
gclssvglx: Thanks @leenagupte 👍  Outdated - resolved.

spec/helpers/check_answers_helper_spec.rb
gclssvglx: Thanks @leenagupte 👍  Outdated - resolved.

spec/helpers/check_answers_helper_spec.rb
gclssvglx: Thanks @leenagupte 👍  Outdated - resolved.

spec/helpers/check_answers_helper_spec.rb
gclssvglx: Thanks @leenagupte 👍  Outdated - resolved.

spec/helpers/check_answers_helper_spec.rb
gclssvglx: Thanks @leenagupte 👍  Outdated - resolved.

app/views/coronavirus_form/check_your_answers/helpers/_list.html.erb
gclssvglx: Thanks @leenagupte 👍  Outdated - resolved.

app/helpers/check_answers_helper.rb
gclssvglx: Thanks @leenagupte 👍  Outdated - resolved.

app/helpers/check_answers_helper.rb
gclssvglx: Thanks @leenagupte 👍  Outdated - resolved.

app/helpers/check_answers_helper.rb
gclssvglx: Thanks @leenagupte 👍  Outdated - resolved.

app/helpers/check_answers_helper.rb
gclssvglx: Thanks @leenagupte 👍  Outdated - resolved.

config/initializers/inflections.rb
gclssvglx: Thanks @leenagupte 👍  Yeah, odd!

spec/helpers/check_answers_helper_spec.rb
gclssvglx: Thanks @leenagupte 👍

spec/helpers/check_answers_helper_spec.rb
gclssvglx: Thanks @leenagupte 👍  Outdated - resolved.

spec/helpers/check_answers_helper_spec.rb
gclssvglx: Thanks @leenagupte 👍  Great suggestion - thanks.  Code updated.

app/views/coronavirus_form/check_your_answers/sections/_space.html.erb
```diff
@@ -0,0 +1,21 @@
...
+} %>
+
+<% unless session[:offer_space] == I18n.t("coronavirus_form.questions.offer_space.options.option_no.label") %>
+  <%= render "govuk_publishing_components/components/summary_list", {
+    title: t("coronavirus_form.check_your_answers.details"),
+    heading_level: 3,
+    heading_size: "m",
+    items: space_items,
+    edit: {
+      href: "offer-space?change-answer",
```
Rosa-Fox: This should be `offer-space-type?change-answer`

Rosa-Fox (changes requested):
Awesome job on this @gclssvglx :)

I believe all of Leena's comments have been addressed. I have tested out the changes and only found something very small that I commented below! After that, LGTM!

app/views/coronavirus_form/check_your_answers/sections/_space.html.erb
gclssvglx: Whoops!  Great spot - thanks @Rosa-Fox 👍

Rosa-Fox (approved):

PR #395, opened 2020-06-19
Fixes double pageview when cookies are accepted

injms:
What & why
----
Two pageviews were being fired when a user accepted cookies using the cookie banner. This is not a good thing, as it artificially inflates views on any page where a user accepts cookies.

The cookie banner was being initalised twice, with two sets of listeners set up to fire pageviews attached to the banner. This pull request removes one of those, so the pageviews now only fire once when a user has accepted cookies.

How to review
-------------

1. Clear all cookies 
1. Check that no tracking blocking plugin is running
1. Check that a GA debugger is running
1. Visit the site
1. Accept cookies

Expected - only one page view is fired once cookies have been accepted.

Links
-----

Trello card: https://trello.com/c/qcLLrnqK

### DIFF
--- a/app/assets/javascripts/analytics.js.erb
+++ b/app/assets/javascripts/analytics.js.erb
@@ -1,8 +1,3 @@
-// Start cookie banner (display settings controlled internally by cookies)
-var element = document.querySelector('[data-module="cookie-banner"]')
-
-new GOVUK.Modules.CookieBanner().start($(element))
-
 var analyticsInit = function() {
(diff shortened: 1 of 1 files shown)


alex-ju (approved):
Tested this locally and works fine. It also fixes a hard error that we currently get on Business Volunteering. I suspect we actually had no pageviews on this service and double on others perhaps. This change is certainly the way forward – no way to write functional tests to cover this change in the current setup, I'm afraid, but something to discuss with the tech leads and potentially investigate further.

PR #400, opened 2020-06-22
Update query string redaction to redact all values

injms:
What
----
All query string values are now redacted, regardless of what they are.

Previously, only two set keys were redacted. This would mean that any change in the keys - eg `product_id` to `productID` - would not be redacted and could cause a leak into Google Analytics. Redacting all the values will prevent this.

This brings this app to parity with alphagov/govuk-coronavirus-vulnerable-people-form/pull/410

## Why

Because PII in GA is bad and this will help prevent any PII from being reported.

How to review

Links

Trello

How to review
-------------

 - Using a GA debugger, check that any URL with query string correctly redacts the values.
 - The Google doc linked to the Trello card may also be useful for giving some more context
 - Check the Jasmine tests cover all bases

Links
-----
 Part of https://trello.com/c/JFMD3vzV

### DIFF
--- a/spec/javascripts/stripPII_spec.js
+++ b/spec/javascripts/stripPII_spec.js
@@ -1,50 +1,39 @@
+/* eslint-env jasmine */
+/* global stripPII */
+
 describe('stripPII', function () {
-  it ('returns an empty string given an empty string', function () {
+  it('returns an empty string given an empty string', function () {
     expect(stripPII('')).toEqual('')
   })
 
-  it ('returns an empty string given a null', function () {
+  it('returns an empty string given a null', function () {
     expect(stripPII(null)).toEqual('')
   })
 
-  it ('replaces nothing if there is no query string', function () {
+  it("does nothing if there's no query string", function () {
     var givenURL = new URL('http://example.com/')
     var expectedURL = 'http://example.com/'
 
     expect(stripPII(givenURL)).toEqual(expectedURL)
   })
 
-  it ('replaces nothing if there are no redactable params', function () {
+  it('redacts a single parameter', function () {
     var givenURL = new URL('http://example.com/?test=blah')
-    var expectedURL = 'http://example.com/?test=blah'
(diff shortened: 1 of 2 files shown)


spec/javascripts/stripPII_spec.js
```diff
@@ -1,50 +1,53 @@
...
+  it('redacts a single parameter', function () {
     var givenURL = new URL('http://example.com/?test=blah')
-    var expectedURL = 'http://example.com/?test=blah'
+    var expectedURL = 'http://example.com/?test=<TEST>'
 
     expect(stripPII(givenURL)).toEqual(expectedURL)
   })
 
-  it ('replaces reference_number', function() {
+  it('redacts reference_number', function () {
```
gclssvglx: I don't think this test is needed any more as the overall functionality is covered by the `redacts all parameters` test below.

spec/javascripts/stripPII_spec.js
```diff
@@ -1,50 +1,53 @@
...
     var givenURL = new URL('http://example.com/?test=blah&reference_number=12345')
-    var expectedURL = 'http://example.com/?test=blah&reference_number=<REFERENCE_NUMBER>'
+    var expectedURL = 'http://example.com/?test=<TEST>&reference_number=<REFERENCE_NUMBER>'
 
     expect(stripPII(givenURL)).toEqual(expectedURL)
   })
 
-  it ('replaces product_id', function() {
+  it('redacts product_id', function () {
```
gclssvglx: Same as `redacts reference_number` - I don't think this is needed anymore 👍

gclssvglx (approved):
Great job 👍

spec/javascripts/stripPII_spec.js
injms: Good point! I've now removed it.

gclssvglx (approved):
LGTM 👍

PR #414, opened 2020-06-30
Remove medical equipment flow

gclssvglx:
What
-----

Update the Business Volunteering form to reflect the fact that we are resuming normal competitive procurement processes for PPE.

Why
-----

As things become more normal, Cabinet Office is moving back to a normal competitive procurement process.

"That means suppliers won't be going to coordination centres, or the PPE Buying Cell, or any other emergency route. They'll be submitting tenders to supply in the usual way."

Links
-----

[Trello card](https://trello.com/c/3VBO6Kbr/652-remove-medical-equipment-flow)


config/routes.rb
```diff
@@ -8,21 +8,10 @@
...
-    # Question 1.0: Can you offer medical equipment?
-    get "/medical-equipment", to: "medical_equipment#show"
-    post "/medical-equipment", to: "medical_equipment#submit"
-
-    # Question 1.2: Tell us about the medical equipment you can offer
-    get "/medical-equipment-type", to: "medical_equipment_type#show"
-    post "/medical-equipment-type", to: "medical_equipment_type#submit"
-
-    # Question 1.4: Testing equipment
-    get "/testing-equipment", to: "testing_equipment#show"
```
steveholmes, line 24: As part of the A-AB-B deployment, we should probably redirect these to the accommodation question, so that anyone mid-form does not get directed to a deleted route.

config/schemas/form_response.json
```diff
@@ -14,16 +13,6 @@
...
-    "medical_equipment": {
-      "$ref": "#/definitions/yes_no"
-    },
-    "medical_equipment_type": {
-      "type": "string",
-      "enum": [
-        "Personal protective equipment (PPE)",
-        "Coronavirus tests or testing facilities"
-      ]
-    },
```
steveholmes: Again, as part of the A-AB-B deployment, should these not be deleted until everyone who already has these in the session submits their answers, so we don't trigger a schema error when submitting?

config/routes.rb
gclssvglx: Good call - thanks @steveholmes 👍

config/schemas/form_response.json
gclssvglx: Good call - thanks @steveholmes 👍

app/helpers/check_answers_helper.rb
```diff
@@ -1,7 +1,5 @@
 module CheckAnswersHelper
   EXCLUDED_QUESTIONS = %w[
```
leenagupte: I think this whole array can be removed.

config/routes.rb
```diff
@@ -8,20 +8,16 @@
...
+    # Redirects for A-AB-B deployment...
+    get "/medical-equipment", to: "accommodation#show"
+    post "/medical-equipment", to: "accommodation#show"
```
leenagupte: Will POSTing to a different controller work? Also, it's POSTING to a GET method.

Perhaps a `redirect` will work better? See: https://github.com/alphagov/govuk-coronavirus-business-volunteer-form/commit/826c290362552e5ebcb14b8539f200d0c30f6067#diff-21497849d8f00507c9c8dcaf6288b136

leenagupte (commented):
Added some initial comments about what jumped out at me. Also, there are some failing tests.

config/routes.rb
gclssvglx: Whoops!  Sorry I'm an idiot.  Thanks 👍

app/helpers/check_answers_helper.rb
gclssvglx: Yup, quite right - thanks @leenagupte 👍

gclssvglx:
Thanks for the comments and suggestions @leenagupte and @steveholmes 👍

gclssvglx:
> Also, there are some failing tests.

Failing tests now fixed

app/controllers/coronavirus_form/accommodation_controller.rb
```diff
@@ -38,16 +40,7 @@ def update_session_store
...
-    if session[:medical_equipment] == I18n.t("coronavirus_form.questions.medical_equipment.options.option_no.label")
-      medical_equipment_url
-    elsif session[:medical_equipment_type] == I18n.t("coronavirus_form.questions.medical_equipment_type.options.number_ppe.label")
-      coordination_centres_url
-    elsif session[:medical_equipment_type] == I18n.t("coronavirus_form.questions.medical_equipment_type.options.number_testing_equipment.label")
-      testing_equipment_url
-    else
-      medical_equipment_url
-    end
+    return check_your_answers_url if session["check_answers_seen"]
```
leenagupte: I think `"/"` should be added as the default previous path for this page, so that users are redirected to the service start page if they click "back"

config/schemas/form_response.json
```diff
@@ -15,15 +14,15 @@
...
+       "$ref": "#/definitions/yes_no"
+     },
+     "medical_equipment_type": {
+       "type": "string",
+       "enum": [
+         "Personal protective equipment (PPE)",
+         "Coronavirus tests or testing facilities"
+       ]
+     },
```
leenagupte: I don't think this indentation is needed, but it's not a blocker.

leenagupte (commented):
There's one minor comment, but otherwise it looks good!

app/controllers/coronavirus_form/accommodation_controller.rb
gclssvglx: Thanks for picking this up @leenagupte 👍  - fixed now

config/schemas/form_response.json
gclssvglx: My bad - copy-and-paste error :-( - good spot thanks 👍

leenagupte (approved):
👍

PR #420, opened 2020-07-13
Bug fix - data not being cleared when going from a 'Yes' option to a 'No' option

gclssvglx:
What
----

When going from a 'Yes' option to a 'No' option on some pages, the old values and data was not being cleared correctly.

This change resolves the issue.

Links
-----

[Trello cards](https://trello.com/c/eGsOEifg/488-update-check-your-answers-page)

### DIFF
--- a/spec/controllers/coronavirus_form/offer_care_controller_spec.rb
+++ b/spec/controllers/coronavirus_form/offer_care_controller_spec.rb
@@ -55,11 +55,17 @@ RSpec.describe CoronavirusForm::OfferCareController, type: :controller do
       expect(response).to redirect_to(offer_other_support_path)
     end
 
-    it "clears previously entered care cost for a 'No' response" do
+    it "clears previously entered answers given a 'No' response" do
+      session[:offer_care_type] = I18n.t("coronavirus_form.questions.offer_care_qualifications.offer_care_type.options").map { |_, item| item[:label] }.sample
+      session[:offer_care_qualifications] = "Gludwch destun ar hap yma."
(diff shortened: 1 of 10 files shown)


reggieb (dismissed):
LGTM

leenagupte (commented):
The code looks good! 
Are there any tests that can be added?

gclssvglx:
Tests added.

spec/controllers/coronavirus_form/offer_care_controller_spec.rb
```diff
@@ -55,11 +55,17 @@
       expect(response).to redirect_to(offer_other_support_path)
     end
 
-    it "clears previously entered care cost for a 'No' response" do
+    it "clears previously entered answers given a 'No' response" do
+      session[:offer_care_type] = I18n.t("coronavirus_form.questions.offer_care_qualifications.offer_care_type.options").map { |_, item| item[:label] }.sample
```
reggieb, line 59: I'd suggest a better pattern to `.map { |_, item| .... }` is `.values.map { |item| }`

reggieb (approved):
Thank you for adding the specs.

spec/controllers/coronavirus_form/offer_care_controller_spec.rb
gclssvglx: Thanks @reggieb - good spot 👍  

I agree, just mapping the values makes more sense than defining a useless and subsequently ignored variable.  However, as the `.map { |_, item| .... }` pattern is used throughout the form, so I'll go with the `prior art` approach.

PR #428, opened 2020-07-24
Remove smoke test prod

GDSNewt:
We need to remove the smoke tests from the Business Volunteer pipeline. This is part of the plan to shut down the service.

We only want to remove the job from the pipeline, not any of the code from the tests, this is because we may need to bring the application back at short notice.

[Trello card](https://trello.com/c/BMd9fzVb/709-remove-the-smoke-tests-from-the-business-volunteer-pipeline)


Rosa-Fox:
@GDSNewt this looks great :) However we need to make sure we don't merge it until after the redirect has been put in place, hence why not approving it yet.

JonathanHallam (approved):
LGTM

